PR #1, opened 2019-06-24
tree: complete the rename to fedora-coreos-koji-tagger

jlebon:
Just something I noticed while perusing the codebase.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ WORKDIR /work
 RUN sed -e "s/[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}/$(uuidgen)/g" /etc/fedora-messaging/fedora.toml > /work/my_config.toml
 
 # Set the Application Name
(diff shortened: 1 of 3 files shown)


dustymabe:
wow. I guess since I created the repo there was some setting where I had notifications disabled. Sorry I didn't see this until now.

Since we moved the repo to https://github.com/coreos/fedora-coreos-releng-automation and the [infra review](https://github.com/dustymabe/fedora-infra-ansible/pull/1#pullrequestreview-254671061) actually asked me to rename it back to just coreos-koji-tagger maybe we should not do this? 

I will definitely update the fedora-messaging config to have the right git repo URL though.

jlebon:
Yup, sure WFM

dustymabe:
> wow. I guess since I created the repo there was some setting where I had notifications disabled. Sorry I didn't see this until now.
> 
> Since we moved the repo to https://github.com/coreos/fedora-coreos-releng-automation and the [infra review](https://github.com/dustymabe/fedora-infra-ansible/pull/1#pullrequestreview-254671061) actually asked me to rename it back to just coreos-koji-tagger maybe we should not do this?
> 
> I will definitely update the fedora-messaging config to have the right git repo URL though.

fixed now: 

https://github.com/coreos/fedora-coreos-releng-automation/blob/681d6b0753da7701e98dffb523d41f0fc21ba643/coreos-koji-tagger/Dockerfile#L26

PR #3, opened 2019-07-05
tree: add new config-bot service

jlebon:
config-bot performs automated management tasks on the main
[fedora-coreos-config](https://github.com/coreos/fedora-coreos-config) repo.

See `README.md` for details.

### DIFF
--- a/config-bot/README.md
+++ b/config-bot/README.md
@@ -1,0 +1,41 @@
+# config-bot
+
+config-bot performs automated management tasks on the main
+[fedora-coreos-config](https://github.com/coreos/fedora-coreos-config)
+repo.
+
+It performs three closely related, but mostly independent
+functions. The names below reflect each of those functions.
+One can find their corresponding configuration section in
+`config.toml` and function names in `main`.
+
+1. `sync-build-lockfiles`: watches for new builds on a set
+   of streams, and pushes the generated lockfile to their
+   corresponding branches
+2. `promote-lockfiles`: on some configurable interval,
+   pushes lockfiles from one config branch to another
+3. `propagate-files`: watches for pushes to a config branch
+   and propagates changes to a subset of files to a set of
+   target branches.
+
+All these are currently timer-based. In the future, they
+will use fedora-messaging for triggering.
+
+Similarly, for now all changes are done using `git push`. In
+the future, config-bot PR functionality can be added.
+
+## Testing locally
+
+Tweak settings as appropriate in `config.toml`, e.g.:
+- point to your fork of `fedora-coreos-config`
+- use your own GitHub token
+- you can comment out whole sections if you'd like to test a
+  specific function only; e.g. if you only want to test
+  `promote-lockfiles`, you can comment out
+  `sync-build-lockfiles` and `propagate-files`
+
+Then:
+
+```
+./main --config myconfig.toml
+```
--- a/config-bot/config.toml
+++ b/config-bot/config.toml
@@ -15,12 +15,12 @@ trigger.mode = 'periodic'
 trigger.period = '15m'
 method = 'push'
 
-[promote-lockfiles]
-source-ref = 'bodhi-updates'
-target-ref = 'testing-devel'
-trigger.mode = 'periodic'
-trigger.period = '24h'
-method = 'push'
+#[promote-lockfiles]
+#source-ref = 'bodhi-updates'
+#target-ref = 'testing-devel'
+#trigger.mode = 'periodic'
+#trigger.period = '24h'
+#method = 'push'
 
 [propagate-files]
 source-ref = 'testing-devel'
--- a/config-bot/config.toml
+++ b/config-bot/config.toml
@@ -1,0 +1,36 @@
+[git]
+author.name = 'CoreOS Bot'
+author.email = 'coreosbot@fedoraproject.org'
+github.repo.owner = 'coreos'
+github.repo.name = 'fedora-coreos-config'
+github.token.username = 'coreosbot'
+github.token.path = '/var/run/secrets/coreos.fedoraproject.org/github-token/token'
+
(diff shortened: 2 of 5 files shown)


config-bot/main
```diff
@@ -0,0 +1,373 @@
...
+        # this allows us to safely use `continue` afterwards
+        if not first:
+            await asyncio.sleep(period)
+        first = False
+
+        for stream in cfg['streams']:
+            builds_json = f'{base_url}/{stream}/builds/builds.json'
+            builds = json.loads(await http_fetch(builds_json))
+            if builds is None:
+                # problem fetching the json; just ignore and we'll retry
```
arithx: Should we add a retry counter and at least print out (if not directly failing) if an individual streams data cannot be retrieved within x tries? With the current setup it could be easy to set it up with a typo in one of the stream names and have the lockfiles never synced.

config-bot/config.toml
```diff
@@ -0,0 +1,36 @@
...
+builds-base-url = 'https://builds.coreos.fedoraproject.org/prod/streams'
+streams = [
+    'bodhi-updates',
+]
+trigger.mode = 'periodic'
+trigger.period = '15m'
+method = 'push'
+
+[promote-lockfiles]
+source-ref = 'bodhi-updates'
```
arithx: Should the `source-ref` & `target-ref` be bundled together in an array to support handling multiple stream combinations at the same time or is the idea to run one instance of this tool per lockfile promotion required?

config-bot/main
```diff
@@ -0,0 +1,373 @@
...
+            if last_source_ref_checksum == source_ref_checksum:
+                continue
+
+            git.checkout(target_ref)
+
+            # get the list of lockfiles from the source ref
+            all_files = git.cmd_output('ls-tree', source_ref,
+                                       '--name-only').splitlines()
+            locks = [f for f in all_files
+                     if fnmatch.fnmatch(f, 'manifest-lock.generated.*.json')]
```
arithx: Minor: could use the `matches_patterns` helper.

config-bot/main
jlebon: > With the current setup it could be easy to set it up with a typo in one of the stream names and have the lockfiles never synced.

Yeah, I think metrics/alerts of some kind are next so we can actually monitor if the app is running as intended without doing `oc logs`. Could probably start with a basic webpage and e.g. emailing if there are X errors within T? Would be cool to plug into Prometheus as well. Let's split out that discussion into a separate ticket which encompasses coreos-koji-tagger too?

> Should we add a retry counter and at least print out (if not directly failing)

Right now, we print on every failure in `http_fetch`. Because this is a long-running app, I'm not sure I want to stop retrying without some kind of higher-level monitoring & alert in place first.

config-bot/config.toml
jlebon: Yeah, I think eventually we'll want to support multiple mappings here, e.g. when `branched -> next-devel` is active for example. Punted on this for now. :) But yeah I think we can handle it all in the same instance.

config-bot/main
```diff
@@ -0,0 +1,373 @@
...
+    for pattern in patterns:
+        if fnmatch.fnmatch(fn, pattern):
+            return True
+    return False
+
+
+# normalize to seconds
+def period_to_seconds(s):
+    assert re.match('^[0-9]+[smh]$', s)
+    n = int(s[:-1])
```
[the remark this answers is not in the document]
jlebon: Thanks for the suggestions, switched to Option 1! (This code was actually copy-pasted from [older code](https://github.com/projectatomic/papr/blob/4b4ad707eed12f801f3f3abe7003ccc159adb3f1/papr/utils.py#L27) I wrote).

Option 2 is neat, though I'm not sure about the added complexity given that we can still express the same durations with Option 1.

config-bot/main
jlebon: Done!

config-bot/main
arithx: > Could probably start with a basic webpage and e.g. emailing if there are X errors within T? Would be cool to plug into Prometheus as well. 

Yeah, having it in Prometheus would make the emailing on x errors or webpage very easy.

>  Because this is a long-running app, I'm not sure I want to stop retrying without some kind of higher-level monitoring & alert in place first.

That's fair.

config-bot/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Testing locally
+
+Tweak settings as appropriate in `config.toml`, e.g.:
+- point to your fork of `fedora-coreos-config`
+- use your own GitHub token
+- you can comment out whole sections if you'd like to test a
+  specific function only; e.g. if you only want to test
+  `promote-lockfiles`, you can comment out
+  `sync-build-lockfiles` and `propagate-files`
```
arithx: RFE suggestion for the future: add flags to the argparse to allow for enabling/disabling or running specific combinations of the commands.

config-bot/main
jlebon: OK, split this out in https://github.com/coreos/fedora-coreos-releng-automation/issues/5

config-bot/README.md
jlebon: https://github.com/coreos/fedora-coreos-releng-automation/issues/6

jlebon:
So, before merging this, there's one major piece left, which is support for lockfile overrides. The reason is that once this is merged and set up, it'll sync the lockfiles from `bodhi-updates` to `testing-devel`, but right now there are a bunch of packages from `coreos-pool` that we want.

I'm working on adding support for overrides in rpm-ostree.

jlebon:
Actually, we could just disable `promote-lockfiles` for now and merge this in so at least we'll have the other functions running.

jlebon:
OK, disabled for now. :arrow_up: 
Merging!

PR #4, opened 2019-07-05
coreos-meta-translator: initial commit

arithx:
Originally discussed in https://github.com/arithx/meta-transpiler/pull/1

Tested against both `plume` & the `fedora-coreos-stream-generator`

### DIFF
--- a/coreos-meta-translator/trans.py
+++ b/coreos-meta-translator/trans.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python3
+
+from argparse import ArgumentParser, FileType
+from datetime import datetime
+import json
+import os
+
+
+def ensure_dup(inp, out, inp_key, out_key):
+    '''
+    If the out dictionary does not contain a value for out_key update it
+    to be equal to the inp dictionaries inp_key value, if it does exist
+    ensure the values are equal between the two dictionaries
+    '''
+    if out.get(out_key, None) == None:
+        out[out_key] = inp.get(inp_key)
+    if out.get(out_key) != inp.get(inp_key):
+        raise Exception("Input Files do not appear to be for the same release")
+
+def url_builder(stream, version, arch, path):
+    return f"https://builds.coreos.fedoraproject.org/prod/streams/{stream}/builds/{version}/{arch}/{path}"
+
+def get_extension(path, modifier):
+    return path.rsplit(modifier)[1][1:]
+
+
+parser = ArgumentParser()
+parser.add_argument("--workdir", help="cosa workdir", required=True)
+parser.add_argument("--build-id", help="build id", required=False)
+args = parser.parse_args()
+
+archs = []
+
+with open(os.path.join(args.workdir, "builds", "builds.json"), 'r') as build_file:
+    build_json = json.load(build_file)
+    if len(build_json.get('builds', [])) > 0:
+        individual_build = {}
+        if args.build_id is None:
+            individual_build = build_json.get('builds')[0]
+            args.build_id = individual_build.get('id')
+        else:
+            for build in build_json.get('builds'):
+                if build.get('id') == args.build_id:
+                    individual_build = build
+                    break
+        archs = individual_build.get('archs')
+
+outer_dir = os.path.join(args.workdir, "builds", args.build_id)
+release_file = os.path.join(outer_dir, "release.json")
+
+out = {}
+if os.path.exists(release_file):
+    with open(release_file, 'r') as w:
+        out = json.load(w)
+
+files = [os.path.join(outer_dir, arch, "meta.json") for arch in archs]
+
+for f in files:
+    with open(f, 'r') as w:
+        input_ = json.load(w)
+
+        arch = input_.get("coreos-assembler.basearch")
+
+        ensure_dup(input_, out, "buildid", "release")
+        ensure_dup(input_.get('coreos-assembler.container-config-git'), out, 'branch', 'stream')
+
+        # build the architectures dict
+        arch_dict = {"media": {}}
+        ensure_dup(input_, arch_dict, "ostree-commit", "commit")
+        generic_arches = ["aws", "qemu", "metal", "openstack", "vmware"]
+        for ga in generic_arches:
+            if input_.get("images", {}).get(ga, None) is not None:
+                i = input_.get("images").get(ga)
+                ext = get_extension(i.get('path'), ga)
+                arch_dict['media'][ga] = {
+                    "artifacts": {
+                        ext: {
+                            "disk": {
+                                "location": url_builder(out.get('stream'), out.get('release'), arch, i.get('path')),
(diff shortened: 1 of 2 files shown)


arithx:
NOTE: tested using expected directory structure while waiting for https://github.com/coreos/coreos-assembler/pull/580 to land

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python3
+
+from argparse import ArgumentParser, FileType
+from datetime import datetime
+import json
+import os
+
+
+def ensureDup(inp, out, inp_key, out_key):
```
jlebon: Minor/optional: `ensure_dup` to match the naming convention?

jlebon: Also, can you add a short docstring describing what this function does?

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+    if out.get(out_key, None) == None:
+        out[out_key] = inp.get(inp_key)
+        return out
```
jlebon: Hmm, shouldn't have to return the `dict` back. Objects are passed by reference in Python.

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+    if out.get(out_key) != inp.get(inp_key):
+        print("Input Files do not appear to be for the same release")
+        raise SystemExit(1)
```
jlebon: I'd just `raise Exception(...)` here.

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+    return out
+
+def url_builder(stream, version, arch, path):
+    return f"https://fcos-builds.s3.amazonaws.com/prod/streams/{stream}/builds/{version}/{arch}/{path}"
```
jlebon: In general, I don't think we should use the `amazonaws.com` at all (and instead `https://builds.coreos.fedoraproject.org/`). Both for encapsulation, but also because `builds.c.fp.o` goes through CloudFront.

lucab: Can this point to the non-AWS URL for the bucket?

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+
+        arch = input_.get("coreos-assembler.basearch")
+
+        out = ensureDup(input_, out, "buildid", "release")
+        out = ensureDup(input_.get('coreos-assembler.container-config-git'), out, 'branch', 'stream')
+
+        # build the architectures dict
+        arch_dict = {"media": {}}
+        arch_dict = ensureDup(input_, arch_dict, "ostree-commit", "commit")
+        generic_arches = [("aws", 2), ("qemu", 2), ("metal", 2), ("openstack", 2), ("vmware", 1)]
```
jlebon: Hmm, I wonder if we could just search for the first `.` instead? That way we don't have to encode these numbers here.

coreos-meta-translator/trans.py
arithx: Do we have a redirection URL to use? If so I can update here & in `plume`.

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            out_arch = out['architectures'][arch]
+            for media_type, val in arch_dict.get('media').items():
+                if media_type not in out_arch['media']:
+                    out['architectures'][arch]['media'].update({media_type: val})
+                elif val == out_arch['media'][media_type]:
+                    continue
+                else:
+                    print("differing media type detected: input_file '{}', media_type '{}'".format(input_file, media_type))
+                    raise SystemExit(1)
```
jlebon: Here too, my suggestion is `raise Exception(...)`.

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+outer_dir = os.path.join(args.workdir, "builds", args.build_id)
+release_file = os.path.join(outer_dir, "release.json")
+
+out = {}
+if os.path.exists(release_file):
+    with open(release_file, 'r') as w:
+        out = json.load(w)
+
+build_dirs = [os.path.join(outer_dir, arch_dir) for arch_dir in os.listdir(outer_dir) if os.path.isdir(os.path.join(outer_dir, arch_dir))]
+files = [os.path.join(build_dir, "meta.json") for build_dir in build_dirs if os.path.exists(os.path.join(build_dir, "meta.json"))]
```
jlebon: Another (somewhat more correct) approach is to get the list of architectures for which the build exists by checking `archs`.

coreos-meta-translator/trans.py
```diff
@@ -0,0 +1,126 @@
...
+def url_builder(stream, version, arch, path):
+    return f"https://fcos-builds.s3.amazonaws.com/prod/streams/{stream}/builds/{version}/{arch}/{path}"
+
+def get_extension(path, ext_count):
+    return ".".join(path.split(".")[-ext_count:])
+
+
+parser = ArgumentParser()
+parser.add_argument("--workdir", help="cosa workdir", required=True)
+parser.add_argument("--build-id", help="build id", required=True)
```
jlebon: Optional: could make this optional so we just default to the last build (first build id in `builds.json`).

coreos-meta-translator/trans.py
jlebon: You should be able to do just `s/fcos-builds.s3.amazonaws.com/builds.coreos.fedoraproject.org/`.

coreos-meta-translator/trans.py
arithx: With the current version scheme we're using `.` in the buildid (though that might be subject to change.

jlebon: Ahhh right! Hmm, OK how about just using the identifier for each artifact (e.g. `qemu`, `openstack`, ...) and get the extension after that? E.g.:

```
In [20]: x = 'fedora-coreos-30.340-qemu.qcow2.gz'

In [21]: x[x.index('qemu')+len('qemu'):]
Out[21]: '.qcow2.gz'
```

coreos-meta-translator/trans.py
arithx: Where does the `archs` key exist?

jlebon: Check out https://github.com/coreos/coreos-assembler/pull/580#issue-292980446.

coreos-meta-translator/trans.py
arithx: Yeah, I can do that (I'll probably just go with an `rstrip` and anything that doesn't fit that mold can have it's own case [like the `installer` or `initramfs`])

arithx:
Updates pushed addressing comments

arithx:
@jlebon squashed

jlebon:
Next step is getting this integrated into the pipeline!

PR #13, opened 2019-07-19
config-bot: adapt for new build layout

jlebon:
We need to teach config-bot to read from the new build layout format
with support for multi-arch. Prep for enabling lockfiles.


jlebon:
Can I get a review on this? @arithx or @lucab maybe? :)

PR #14, opened 2019-07-22
coreos-koji-tagger: add target release heuristic for fedora-coreos-pi…

rfairley:
…nger

Like `zincati` this is built as a module - use a similar heuristic.

---

This has not been tested, however this package is built in a similar way to `zincati` and should work with the same conditions.


rfairley:
Required by: https://pagure.io/dusty/coreos-koji-data/pull-request/8

jlebon:
Thanks! (Though yeah, we should generalize this soon...)

PR #17, opened 2019-07-26
coreos-koji-tagger: point at fedora-coreos-config repo

jlebon:
Start listening for git pushes on
https://github.com/coreos/fedora-coreos-config instead of
https://pagure.io/dusty/coreos-koji-data.

This notably involves switching out the topic we listen to, and
introspecting the message. We could use the `commit[].modified[]` values
here to be even more selective, but leaving that for a follow-up.

### DIFF
--- a/coreos-koji-tagger/Dockerfile
+++ b/coreos-koji-tagger/Dockerfile
@@ -31,8 +31,8 @@ RUN sed -i 's/INFO/WARNING/' /work/my_config.toml
 # Set the format for the log messages
 RUN sed -i 's/format =.*$/format = "%(asctime)s %(levelname)s %(name)s - %(message)s"/' /work/my_config.toml
 
-# We only care about pungi.compose.status.change messages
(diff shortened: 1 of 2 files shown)


dustymabe:
looks like you already have the GH repo hooked up to fedmsg/datagrepper so that is nice: https://apps.fedoraproject.org/datagrepper/id?id=2019-e06a979c-6e1c-4f0e-ae97-cbf8b7b9660c&is_raw=true&size=extra-large

coreos-koji-tagger/Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN sed -i 's/INFO/WARNING/' /work/my_config.toml
 RUN sed -i 's/format =.*$/format = "%(asctime)s %(levelname)s %(name)s - %(message)s"/' /work/my_config.toml
 
 # We only care about pungi.compose.status.change messages
```
dustymabe: looks like this comment should be updated too ^^

dustymabe:
mostly LGTM - can we separate whitespace changes into a separate commit?

jlebon:
I just dropped the whitespace changes for now -- those were from my editor.

PR #19, opened 2019-08-12
coreos-koji-tagger: get latest updates in Dockerfile

dustymabe:
I've hit issues where the container was out of date and some things
were slightly broken. One case of this: https://pagure.io/releng/compose-tracker/pull-request/21


mike-nguyen:
LGTM

PR #20, opened 2019-08-12
coreos-koji-tagger: ability to override github repo

dustymabe:
This should make it easier to test in fedora's stage environment.

Also switch to `os.getenv`


dustymabe:
> though I think we should actually just split out a config file with all the knobs there

I think I agree.

PR #22, opened 2019-08-12
coreos-koji-tagger: comment out a few debug statements

dustymabe:
These make the logs a little too verbose.

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -212,8 +212,8 @@ class Consumer(object):
             logger.info('Will not attempt koji write operations')
 
     def __call__(self, message: fedora_messaging.api.Message):
-        logger.debug(message.topic)
-        logger.debug(message.body)
+        #logger.debug(message.topic)
+        #logger.debug(message.body)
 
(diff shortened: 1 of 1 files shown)


coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -212,8 +212,8 @@ def __init__(self):
             logger.info('Will not attempt koji write operations')
 
     def __call__(self, message: fedora_messaging.api.Message):
-        logger.debug(message.topic)
-        logger.debug(message.body)
+        #logger.debug(message.topic)
+        #logger.debug(message.body)
```
mike-nguyen: Is there a reason why these lines aren't deleted rather than commented out?

dustymabe: mostly so they'd be easy to add back without having to remember exactly the syntax (i.e. var name) again. I could be swayed to delete it but have a preference for leaving it in and commented.

mike-nguyen: I can go either way with this also as long as there isn't commented code everywhere which is not the case for this.

PR #23, opened 2019-08-12
coreos-koji-tagger: get rid of bare raise

### LINKED ISSUE #21: [secaudit,tagger,info] Bare raise
Rev: `06275b45306167c7ee86649dc6936bb637138850`
`coreos-koji-tagger/coreos_koji_tagger.py`
Line: 207

Bare raise is used outside of a `try`/`catch` block.
This will trigger a `TypeError`, since Python will try to raise the last raised exception, which will most likely be `None`.

dustymabe:
Raise an Exception object instead of a bare raise.

Fixes #21


dustymabe:
thanks @puiterwijk

PR #28, opened 2019-08-13
Rewrite tagger to use koji Python API rather than shelling out

puiterwijk:
Fixes: #2
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

PR #29, opened 2019-08-14
coreos-koji-tagger: move to koji python library

dustymabe:
This PR will convert us to using the koji python library. There are several
other small commits in this PR that can be reviewed separately. The
commit summary is:

- move to koji python library
- find_principal_from_keytab no longer class function
- convert some log statements to info
- wait for tagging tasks to complete
- no more bare raises
- catch exceptions and continue

### DIFF
--- a/coreos-koji-tagger/Dockerfile
+++ b/coreos-koji-tagger/Dockerfile
@@ -10,6 +10,12 @@ RUN dnf update -y && dnf clean all
 # Install pagure/fedmsg libraries
 RUN dnf -y install dnf-plugins-core python3-libpagure fedora-messaging koji krb5-workstation && dnf clean all
 
+# Update koji src to 1.18 for multicall support (not yet packaged as an RPM)
+RUN dnf -y install git && dnf clean all
+RUN rm -rf /usr/lib/python3.7/site-packages/koji
+RUN git -C /opt/ clone -b koji-1.18.0 https://pagure.io/koji.git
+ENV PYTHONPATH=/opt/koji/
+
 # Grab the kerberos/koji configuration (i.e. /usr/bin/stg-koji) by
 # installing the fedora-packager rpm. We don't need the deps because
 # we aren't building anything.
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -16,7 +16,7 @@ import requests
 
 # Set local logging 
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
+logger.setLevel(logging.INFO)
 
 # The target and the intermediate tag. The target tag is where we want
 # builds to end up. We'll check the target tag to see if builds are already
@@ -241,8 +241,8 @@ class Consumer(object):
             logger.info('Will not attempt koji write operations')
 
     def __call__(self, message: fedora_messaging.api.Message):
-        #logger.debug(message.topic)
-        #logger.debug(message.body)
+        logger.debug(message.topic)
+        logger.debug(message.body)
 
         # Grab the raw message body and the status from that
         msg = message.body
@@ -251,7 +251,7 @@ class Consumer(object):
         commit = msg['head_commit']['id']
 
         if (repo != self.github_repo_fullname):
-            logger.debug(f'Skipping message from unrelated repo: {repo}')
+            logger.info(f'Skipping message from unrelated repo: {repo}')
             return
 
         if (branch != self.github_repo_branch):
@@ -414,12 +414,12 @@ def find_principal_from_keytab(keytab: str) -> str:
 
     # The principal will be the last column in that line
     principal = line.split(' ')[-1]
-    logger.debug(f'Found principal {principal} in keytab')
+    logger.info(f'Found principal {principal} in keytab')
     return principal
 
 def runcmd(cmd: list, **kwargs: int) -> subprocess.CompletedProcess:
     try:
-        logger.debug(f'Running command: {cmd}')
+        logger.info(f'Running command: {cmd}')
         cp = subprocess.run(cmd, **kwargs)
     except subprocess.CalledProcessError as e:
         logger.error('Running command returned bad exitcode')
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -7,6 +7,7 @@ import logging
 import json
 import koji
 from koji_cli.lib import watch_tasks
+import traceback
 
 import dnf.subject
 import hawkey
@@ -242,6 +243,20 @@ class Consumer(object):
             logger.info('Will not attempt koji write operations')
 
     def __call__(self, message: fedora_messaging.api.Message):
+        # Catch any exceptions and don't raise them further because
+        # it will cause /usr/bin/fedora-messaging to crash and we'll
+        # lose the traceback logs from the container
+        try:
+            self.process(message)
+        except Exception as e:
+            logger.error('Caught Exception!')
+            logger.error('###################################')
+            traceback.print_exc()
+            logger.error('###################################')
+            logger.error('\t continuing...')
+            pass
+
+    def process(self, message: fedora_messaging.api.Message):
         logger.debug(message.topic)
         logger.debug(message.body)
 
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -359,7 +359,9 @@ class Consumer(object):
 
 
     def get_buildsinfo_from_rpmnevras(self, rpmnevras: set) -> dict:
-        # Given a set of rpm NEVRAs get a set of corresponding koji buildids
+        """
+        Given a set of rpm NEVRAs get a set of corresponding koji buildids
+        """
         if not rpmnevras:
             raise ValueError("No nevras to get_builds_from_rpmnevras")
 
@@ -401,17 +403,23 @@ class Consumer(object):
         return buildsinfo
 
     def get_pkglist_in_tag(self, tag: str) -> set:
-        # Given a tag, return the set of packages in its pkglist
+        """
+        Given a tag, return the set of packages in its pkglist
+        """
         pkgs = self.koji_client.listPackages(tagID=tag)
         return set([pkg['package_name'] for pkg in pkgs])
 
     def get_tagged_buildids(self, tag: str) -> set:
-        # Given a tag, return the buildids tagged into it
+        """
+        Given a tag, return the buildids tagged into it
+        """
         builds = self.koji_client.listTagged(tag=tag)
         return set([build['build_id'] for build in builds])
 
 def find_principal_from_keytab(keytab: str) -> str:
-    # Find the pricipal/realm that the keytab is for
+    """
+    Find the pricipal/realm that the keytab is for
+    """
     cmd = ['/usr/bin/klist', '-k', keytab]
     cp = runcmd(cmd, capture_output=True, check=True)
 
@@ -466,21 +474,23 @@ def get_NVRA_from_NEVRA(string: str) -> str:
     return nvra
 
 def parse_lockfile_data(text: str) -> list:
-    # Parse the rpm lockfile format and return a list of rpms in
-    # NEVRA form.
-    # Best documention on the format for now:
-    #     https://github.com/projectatomic/rpm-ostree/commit/8ff0ee9c89ecc0540182b5b506455fc275d27a61
-    #
-    # An example looks something like:
-    #
-    #   {
-    #     "packages": {
-    #       "GeoIP": {
-    #         "evra": "1.6.12-5.fc30.x86_64",
-    #         "digest": "sha256:21dc1220cfdacd089c8c8ed9985801a9d09edb7c26543694cef57ada1d8aafa8"
-    #       }
-    #     }
-    #   }
+    """
+    Parse the rpm lockfile format and return a list of rpms in
+    NEVRA form.
+    Best documention on the format for now:
+        https://github.com/projectatomic/rpm-ostree/commit/8ff0ee9c89ecc0540182b5b506455fc275d27a61
+    
+    An example looks something like:
+    
+      {
+        "packages": {
+          "GeoIP": {
+            "evra": "1.6.12-5.fc30.x86_64",
+            "digest": "sha256:21dc1220cfdacd089c8c8ed9985801a9d09edb7c26543694cef57ada1d8aafa8"
+          }
+        }
+      }
+    """
 
     # The data is JSON (yay)
     data = json.loads(text)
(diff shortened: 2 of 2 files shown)


coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -183,9 +183,37 @@
 )
 
 
+class BuildInfo(object):
+    def __init__(self, id, buildrootid):
+        self.buildrootid = buildrootid
+        self.id          = id
```
arithx: `id` is a builtin function as well, we should use something else in this function (leaving the class variable named `id` is fine though), e.x.:

```
def __init__(self, id_, buildrootid):
  self.id = id_
```

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -197,15 +225,16 @@ def __init__(self):
                                         'GITHUB_REPO_BRANCH',
                                         DEFAULT_GITHUB_REPO_BRANCH)
         self.koji_user         = COREOS_KOJI_USER
+        self.koji_client       = koji.ClientSession(KOJI_SERVER_URL)
 
-        # If a keytab was specified let's use it
+        # If a keytab was specified let's try to auth.
         self.keytab_file = os.getenv('COREOS_KOJI_TAGGER_KEYTAB_FILE')
+        if self.keytab_file and not os.path.exists(self.keytab_file):
```
arithx: Any reason to not just embed this check inside the following `if` block? e.x.:

```
if self.keytab_file:
  if not os.path.exists(self.keytab_file):
    raise Exception("...")
  principal = ...
```

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -264,69 +288,111 @@ def __call__(self, message: fedora_messaging.api.Message):
...
-        # Grab the currently tagged builds and convert it into a set
-        currentbuilds = set(get_tagged_builds(self.target_tag))
+        # Get the set of names of pkgs needed
+        desiredpkgs = set([x.name for x in buildsinfo.values()])
 
-        # Find out the difference between the current set of builds
-        # that exist in the koji tag and the desired set of builds to
-        # be added to the koji tag.
-        buildstotag = list(desiredbuilds.difference(currentbuilds))
+        # Grab the set of names of pkgs that can be tagged into the tag
```
arithx: Tagging things into a tag doesn't really make the most sense from the outside perspective but I think that might be outside the scope of this review :)

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -264,69 +288,111 @@ def __call__(self, message: fedora_messaging.api.Message):
...
+            buildsinfo[buildid].releasever = \
+                get_releasever_from_buildroottag(
+                    brinfo[buildid].result['tag_name'])
+
+        return buildsinfo
+
+    def get_pkglist_in_tag(self, tag: str) -> set:
+        # Given a tag, return the set of packages in its pkglist
+        pkgs = self.koji_client.listPackages(tagID=tag)
+        return set([pkg['package_name'] for pkg in pkgs])
```
arithx: Does `koji_client` guarantee that `package_name` will exist? If not we should probably use a safe getter.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -264,69 +288,111 @@ def __call__(self, message: fedora_messaging.api.Message):
...
-                                f'{tag} tag: {buildstotagforthisrelease}')
-                    if self.keytab_file:
-                        tag_builds(tag=tag, builds=buildstotagforthisrelease)
+            # Get a set of tuples of build name to tag to tag into
+            tuples = [(self.intermediate_tag.format(
+                        releasever=buildsinfo[x].releasever),
+                        buildsinfo[x].nvr)
+                        for x in buildstotag]
+            logger.info('Tagging the following (tag, nvr): \n\t%s' %
+                                            '\n\t'.join(map(str, tuples)))
```
arithx: The spacing on this line seems inconsistent.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -215,11 +244,6 @@ def __call__(self, message: fedora_messaging.api.Message):
         #logger.debug(message.topic)
```
arithx: We should clean up these commented debug lines.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -197,15 +225,16 @@ def __init__(self):
...
         if self.keytab_file:
-            if os.path.exists(self.keytab_file):
-                self.kinit()
-            else:
-                raise Exception("The specified keytab file "
-                                "does not exist: %s" % self.keytab_file)
+                principal = self.find_principal_from_keytab()
+                self.koji_client.gssapi_login(principal, self.keytab_file)
         else:
             logger.info('No keytab file defined in '
```
arithx: Is there ever a case where we want to run this without a keytab file being defined & existing?

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -264,69 +288,111 @@ def __call__(self, message: fedora_messaging.api.Message):
...
+                        releasever=buildsinfo[x].releasever),
+                        buildsinfo[x].nvr)
+                        for x in buildstotag]
+            logger.info('Tagging the following (tag, nvr): \n\t%s' %
+                                            '\n\t'.join(map(str, tuples)))
+            if self.keytab_file:
+                with self.koji_client.multicall(strict=True) as m:
+                    for (tag, nvr) in tuples:
+                        m.tagBuild(tag=tag, build=nvr)
             logger.info('Tagging done')
```
arithx: This line feels a bit disingenuous; should it perhaps be moved inside of the `if self.keytab_file` block so that it only prints if an action was taken?

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -264,69 +288,111 @@ def __call__(self, message: fedora_messaging.api.Message):
...
 
+
+    def get_buildsinfo_from_rpmnevras(self, rpmnevras: set) -> dict:
+        # Given a set of rpm NEVRAs get a set of corresponding koji buildids
```
arithx: This feels like it is intended as a docstring, if so let's follow [PEP 257](https://www.python.org/dev/peps/pep-0257/#specification) and switch it to triple double quotes.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -373,27 +429,29 @@ def runcmd(cmd: list, **kwargs: int) -> subprocess.CompletedProcess:
         raise
     return cp # subprocess.CompletedProcess
 
-def get_rich_info_for_rpm_string(string: str, arch: bool) -> hawkey.NEVRA:
-    # arch: (bool) whether arch is included in the string
-    if arch:
+def get_NVRA_from_NEVRA(string: str) -> str:
```
arithx: Everything in this function is double spaced now

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -373,27 +429,29 @@ def runcmd(cmd: list, **kwargs: int) -> subprocess.CompletedProcess:
...
-    #   print(info.epoch)
-    #   print(info.release)
-    #   print(info.arch)
-    return info
+        # get a list of hawkey.NEVRA objects that are the possibilities
+        nevras  = subject.get_nevra_possibilities(forms=form)
+
+        # return the first hawkey.NEVRA item in the list of possibilities
+        rpminfo = nevras[0]
+        #   print(rpminfo.name)
```
arithx: Let's drop these commented out lines.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -336,8 +337,11 @@ def __call__(self, message: fedora_messaging.api.Message):
                                             '\n\t'.join(map(str, tuples)))
             if self.keytab_file:
                 with self.koji_client.multicall(strict=True) as m:
-                    for (tag, nvr) in tuples:
-                        m.tagBuild(tag=tag, build=nvr)
+                    tasks = [m.tagBuild(tag=tag, build=nvr)
+                                    for (tag, nvr) in tuples]
+                watch_tasks(self.koji_client,
+                            [task.result for task in tasks],
```
arithx: Does this work? It roughly reads like `tasks` would be defined out of scope of this line (but I wouldn't be surprised if somehow `with` doesn't change the local scoping).

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: yes. It's usefull for running in a read-only/dry-run mode for testing. The following log line says: `Will not attempt koji write operations`

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: i can merge them

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: fixed in next upload

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: I chose to leave them commented out in a previous commit: see https://github.com/coreos/fedora-coreos-releng-automation/pull/22#discussion_r313078517

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: :)

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: I was mostly trying to carry forward to a new line and then make the second line be longer than the first line. In a future PR I'll run the whole thing through the `black` formatter.

coreos-koji-tagger/coreos_koji_tagger.py
arithx: Since we're bumping the logger level to `INFO` in a later commit is there any reason not to just uncomment them or to change up the comments to describe the fields/variables available on the object rather than just commented out logger lines?

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: yes.. If it wasn't there we'd want an exception

coreos-koji-tagger/Dockerfile
```diff
@@ -10,6 +10,12 @@ RUN dnf update -y && dnf clean all
 # Install pagure/fedmsg libraries
 RUN dnf -y install dnf-plugins-core python3-libpagure fedora-messaging koji krb5-workstation && dnf clean all
 
+# Update koji src to 1.18 for multicall support (not yet packaged as an RPM)
+RUN dnf -y install git && dnf clean all
+RUN rm -rf /usr/lib/python3.7/site-packages/koji
+RUN git -C /opt/ clone https://pagure.io/koji.git
```
jlebon: Hmm, if we're cloning straight from upstream, let's at least clone a specific ref here?

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: fixed

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: fixed in next upload

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: k

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: it works in testing

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: indented!

coreos-koji-tagger/coreos_koji_tagger.py
arithx: Huh, TIL that `with` doesn't change scoping.

coreos-koji-tagger/Dockerfile
dustymabe: done

dustymabe:
ok i've addressed all code review comments. Please do review but let me take some time tomorrow to do some final testing before I merge.

coreos-koji-tagger/Dockerfile
```diff
@@ -10,6 +10,12 @@ RUN dnf update -y && dnf clean all
 # Install pagure/fedmsg libraries
 RUN dnf -y install dnf-plugins-core python3-libpagure fedora-messaging koji krb5-workstation && dnf clean all
 
+# Update koji src to 1.18 for multicall support (not yet packaged as an RPM)
+RUN dnf -y install git && dnf clean all
+RUN rm -rf /usr/lib/python3.7/site-packages/koji
+RUN git -C /opt/ clone https://pagure.io/koji.git && git -C /opt/koji checkout koji-1.18.0
```
jlebon: Minor: that's `-b koji-1.18.0` :)

dustymabe: i'm checking out the `koji-1.18.0` tag. Not trying to create a new branch or anything:

```
Note: checking out 'koji-1.18.0'.                                                                                  
                                                                                                                   
You are in 'detached HEAD' state. You can look around, make experimental                                                                                                                                                              
changes and commit them, and you can discard any commits you make in this                                                                                                                                                             
state without impacting any branches by performing another checkout.                                                                                                                                                                  
                                                                                                                                                                                                                                      
If you want to create a new branch to retain commits you create, you may                                                                                                                                                              
do so (now or later) by using -b with the checkout command again. Example:                                                                                                                                                            
                                                                                                                   
  git checkout -b <new-branch-name>                                                                                
                                                                                                                                                                                                                                      
HEAD is now at c7657331 PR#1588: 1.18 release
```

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: yeah if I find out otherwise tomorrow when I do more tests I'll report back here.

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: since we did change to INFO, I agree. I'll uncomment them in that commit

dustymabe:
ok I think i've addressed all outstanding code review comments

coreos-koji-tagger/Dockerfile
jlebon: I meant `clone -b`, not `checkout -b`.

dustymabe: ok I see. So if I `git clone -b koji-1.18.0` I won't need the `&& git checkout koji-1.18.0`

dustymabe:
> Seems like this obsoletes #28, correct?

correct - added that to the commit message

dustymabe:
ok - performing some local testing now

dustymabe:
> ok - performing some local testing now

Testing looks good! Are there any items left for the review?

dustymabe:
merging since @jlebon is out today.. We can address any outstanding issues in a new PR.

PR #31, opened 2019-08-19
config-bot: enable promote-lockfiles

jlebon:
We're in position to enable lockfile promotion now that
coreos-koji-tagger is set up and monitoring the `testing-devel` branch.
Let's do this!

### DIFF
--- a/config-bot/config.toml
+++ b/config-bot/config.toml
@@ -15,12 +15,12 @@ trigger.mode = 'periodic'
 trigger.period = '15m'
 method = 'push'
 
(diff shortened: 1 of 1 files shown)


jlebon:
Marking as draft for now; want to do a final sanity check on this.

jlebon:
OK, this works well. There's one gotcha, which is that `testing-devel` doesn't have the modular repos enabled right now, so once the lockfile is promoted, some packages will fail to resolve until the coreos-pool repo is respinned (see related issue #32).

PR #33, opened 2019-08-19
coreos-koji-tagger: fix manifest lockfile filename

jlebon:
The lockfile is located at `manifest-lock.x86_64.json` on
`testing-devel`. The `.generated` one was from a test message that was
actually for `bodhi-updates`.

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -134,7 +134,7 @@ EXAMPLE_MESSAGE_BODY = json.loads("""
       "distinct": true,
       "timestamp": "2019-07-26T16:11:15Z",
       "modified": [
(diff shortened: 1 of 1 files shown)


jlebon:
Hold for a sec

jlebon:
OK, this is definitely correct, but the "unit test" (i.e. just running `./coreos_koji_tagger.py`) no longer works (I think after #29). Anyway, I still updated the sample message for when we fix it.

dustymabe:
> TypeError: 'bool' object is not callable

We are taking advantage of some [new code that is only in koji 1.18](https://github.com/coreos/fedora-coreos-releng-automation/blob/1fcfacf9da981f259e23e94a8cdce2432244a207/coreos-koji-tagger/Dockerfile#L13-L17). Do you have 1.18

PR #39, opened 2019-08-29
coreos-koji-tagger: add support for overrides and multi-arch lockfiles

dustymabe:
Add support for the other lockfiles and a few other fixups:

- coreos-koji-tagger: move commit var access lower
- coreos-koji-tagger: Update README with more staging docs
- coreos-koji-tagger: add support for other architectures
- coreos-koji-tagger: support overrides files
- coreos-koji-tagger: remove koji 1.18 workaround

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -283,7 +283,7 @@ class Consumer(object):
         # Now grab lockfile data from the commit we should operate on:
         desiredrpms = set()
         for arch in ['x86_64', 'aarch64', 'ppc64le', 's390x']:
-            for lockfile in ['manifest-lock']:
+            for lockfile in ['manifest-lock', 'manifest-lock.overrides']:
                 url = f'https://raw.githubusercontent.com/{repo}/{commit}/{lockfile}.{arch}.json'
                 logger.info(f'Attempting to retrieve data from {url}')
                 r = requests.get(url)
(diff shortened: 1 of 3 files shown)


coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -271,13 +271,17 @@ def process(self, message: fedora_messaging.api.Message):
...
-        r = requests.get(url)
-
+        desiredrpms = set()
+        for arch in ['x86_64', 'aarch64', 'ppc64le', 's390x']:
+            for lockfile in ['manifest-lock']:
+                url = f'https://raw.githubusercontent.com/{repo}/{commit}/{lockfile}.{arch}.json'
+                logger.info(f'Attempting to retrieve data from {url}')
+                r = requests.get(url)
+                # parse the lockfile and add the set of rpm NEVRAs (strings)
+                if r.ok:
```
cgwalters: I think we should differentiate between `404` and any other error.  At least log non-404.

dustymabe: Yeah Let's log errors. Really even 404 shouldn't happen once we have all arches in place, so let's log that one too.

dustymabe: should be fixed now

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -274,6 +273,7 @@ def process(self, message: fedora_messaging.api.Message):
 
         # Now grab data from the commit we should operate on:
         # XXX: should update for multi-arch
+        commit = msg['head_commit']['id']
```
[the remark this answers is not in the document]
dustymabe: converted it to a safe getter!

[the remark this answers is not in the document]
dustymabe: and then I converted it back.. there are messages with `head_commit` set to null so now I'm checking that first and then pulling `id` afterwards.

PR #42, opened 2019-09-23
config-bot: include source ref checksum used in commit message

jlebon:
That way we know exactly what content from what commit was propagated.
Similarly, in the promotion script, include the commit that was
promoted.

Hat tip to Colin who suggested this during a review of the FCOS release
process.


dustymabe:
This is nice. Maybe in the future we can include the commit ids of all the commits the last merge. `git shortlog --no-merges --pretty=format:"%h %s" -e LASTMERGE..HEAD` would show a nice summary.

jlebon:
I like that; opened https://github.com/coreos/fedora-coreos-releng-automation/issues/44.

PR #45, opened 2019-09-24
Add coreos-ostree-importer

dustymabe:


dustymabe:
Just putting this up here for now. It mostly works but I need to wire up the fedora messaging piece.

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+
+FEDORA_MESSAGING_TOPIC_LISTEN = (
+    "org.fedoraproject.prod.coreos.build.request.ostree-import"
+)
+FEDORA_MESSAGING_TOPIC_RESPOND = FEDORA_MESSAGING_TOPIC_LISTEN + ".finished"
+
+
+# We are processing the org.fedoraproject.prod.coreos.build.request.ostree-import topic
+# https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.coreos.build.request.ostree-import&delta=100000
+EXAMPLE_MESSAGE_BODY = json.loads(
```
jlebon: Probably could throw a link to https://github.com/coreos/fedora-coreos-tracker/issues/198#issuecomment-513944390 here, though will try to throw that up as a PR.

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+# Given a repo (and thus an input JSON) analyze existing koji tag set
+# and tag in any missing packages
+class Consumer(object):
+    def __init__(self):
+        logger.info(
+            "Processing messages with topic: %s" % FEDORA_MESSAGING_TOPIC_LISTEN
+        )
+
+    def __call__(self, message: fedora_messaging.api.Message):
+        # Catch any exceptions and don't raise them further because
```
jlebon: Hmm, I think a better approach here would be to define a new exception type and throw that in our code when checking for error conditions.

This kind of exception handling means that we don't discriminate between e.g. platform errors, programmer errors, and just invalid messages.

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+        sha256sum = checksum[7:]
+        target_repo_path = KNOWN_OSTREE_REPOS[target_repo]
+        source_repo_path = None
+
+        # Parse the message
+
+        # Check the target repo to make sure it exists
+        # and is accessible
+        if not ostree_repo_exists(target_repo_path):
+            raise Exception(f"OSTree repo does not exist at {target_repo_path}")
```
jlebon: Can we just check this upfront at init time?

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+            if target_repo == "prod" and commit_exists(
+                repo=KNOWN_OSTREE_REPOS["compose"], commit=ostree_checksum
+            ):
+                source_repo_path = KNOWN_OSTREE_REPOS["compose"]
+            else:
+                # Grab the file from s3 and then pull local
+                untar_file_from_s3(url=commit_url, tmpdir=tmpdir, sha256sum=sha256sum)
+                source_repo_path = tmpdir
+
+            # one more sanity check: make sure buildid == version
```
jlebon: Nice!

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+            assert_commit_has_version(
```
jlebon: Another sanity check we should do is that the latest commit on the target ref is the parent of the new commit we're importing to make sure we're not introducing any holes.

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+    return cp  # subprocess.CompletedProcess
+
+
+def ostree_pull_local(srcrepo: str, dstrepo: str, branch: str, commit: str):
+    # pull content
+    cmd = ["ostree", f"--repo={dstrepo}", "pull-local", srcrepo, commit]
+    runcmd(cmd)
+    # update branch
+    if ostree_branch_exists(repo=dstrepo, branch=branch):
+        cmd = ["ostree", f"--repo={dstrepo}", "reset", branch, commit]
```
jlebon: See sanity-check related to this up above

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -0,0 +1,280 @@
...
+
+        # Parse the message
+
+        # Check the target repo to make sure it exists
+        # and is accessible
+        if not ostree_repo_exists(target_repo_path):
+            raise Exception(f"OSTree repo does not exist at {target_repo_path}")
+
+        # Detect if the commit already exists in the target repo
+        if commit_exists(target_repo_path, ostree_checksum):
```
jlebon: One caveat here is that this ignores detached commit metadata changes. Which should be fine for the most part... our OSTree commits should already be signed by the time we request it get imported. Though one would think that calling `ostree-import` twice, once without the detached metadata in place and once more with would behave correctly. Probably worth a comment or something for now?

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: will do

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: +1

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: +1

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: Added a comment about the assumption we're making

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: You mean to link to that comment rather than having the `EXAMPLE_MESSAGE_BODY` variable? I `EXAMPLE_MESSAGE_BODY` when testing in the `main()`.

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: This runs in an infinite loop in openshift. What I've found is that it's much easier to have the exceptions get thrown and viewable in the current container's log history than it is to distinguish between exceptions and `raise` or `pass` based on the type. 

If something isn't behaving properly then we can see the logs and fix it, right?

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: will do

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: Let me know if it's good enough!

dustymabe:
Ok addressed some code review comments. Added a README. Also added in Fedora messaging configs. Marking this as ready for review.

coreos-ostree-importer/coreos_ostree_importer.py
jlebon: > You mean to link to that comment rather than having the EXAMPLE_MESSAGE_BODY variable?

No, just adding it as a comment to give a linkback with more context on where the schema comes from.

coreos-ostree-importer/coreos_ostree_importer.py
jlebon: Right, but a service being "up" (or Ready if there's a readiness probe) in Kubernetes should mean that it's available to handle requests. If in fact there's actually something wrong and we're not "passing that up" to Kubernetes, we lose those semantics and e.g. basic telemetry/alerts/`oc get pods` become less useful.

Of course, since we're pretty integrated right now, I agree it doesn't matter too much (if importing fails, we'll just go look at the container logs and see what's up). I tried looking for a good doc on best practices around error-handling for k8s services, but couldn't find anything good. Anyway, overall I don't think this is the best approach, but I'm good with just getting this in as is for now!

dustymabe: I agree that what we're doing here is a bit of an anti-pattern and that errors should be bubbled up and reported at a higher level. AFAIU that higher level reporting/notifications doesn't exist yet (at least for this specific openshift environment) so this is the best I've come up with to make sure I can see the error messages and not lose them. Let's leave this as is for now and modify when the reporting situation is better?

jlebon: Yup, sure.

coreos-ostree-importer/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+cat <<'EOF' > publisher.py
+#!/usr/bin/python3
+from fedora_messaging import api, message
+topic = 'org.fedoraproject.prod.coreos.build.request.ostree-import'
+body = {
+    "build_id": "30.20190905.0",
+    "stream": "testing",
+    "basearch": "x86_64",
+    "commit": "s3://fcos-builds/prod/streams/testing/builds/30.20190905.0/x86_64/ostree-commit.tar",
```
dustymabe: @jlebon any chance we could just make the ACL for this public and use an https url here instead so we don't have to worry about creating read-only credentials for AWS for this service?

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: added a comment with the link!

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: 👍

coreos-ostree-importer/README.md
jlebon: AFAIU, there should already be credentials available to use. See e.g. https://pagure.io/fedora-infrastructure/issue/8189#comment-600735 for a similar concern with robosignatory. Hmm, although that's probably not accessible to OpenShift apps.

If not, sure we can just use `https://`. The object is already public, but I guess I want to keep the door open for uploading bucket objects as private. But yeah... if we really wanted to do this, we could request credentials then.

dustymabe: > AFAIU, there should already be credentials available to use. See e.g. https://pagure.io/fedora-infrastructure/issue/8189#comment-600735 for a similar concern with robosignatory. Hmm, although that's probably not accessible to OpenShift apps.

right. I think I would want a separate set of credentials for this app that were read-only.

> 
> If not, sure we can just use `https://`. The object is already public, but I guess I want to keep the door open for uploading bucket objects as private. But yeah... if we really wanted to do this, we could request credentials then.

It would be nice if we can do `https://` so I don't have to ask for new creds and also it will reduce the complexity of coreos-ostree-importer. Maybe we can do that soonish.

PR #48, opened 2019-10-08
WIP: config-bot: add bump-lockfiles functionality

jlebon:
This is meant to address a smarter way to bump lockfiles.
See: https://github.com/coreos/fedora-coreos-tracker/issues/293

Only supports `git push` for now. Working on making that smarter.

### DIFF
--- a/config-bot/Dockerfile
+++ b/config-bot/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:30
+FROM quay.io/coreos-assembler/coreos-assembler:master
 
 RUN dnf -y install git python3-toml python3-aiohttp && dnf clean all
 
--- a/config-bot/config.toml
+++ b/config-bot/config.toml
@@ -15,11 +15,20 @@ trigger.mode = 'periodic'
 trigger.period = '15m'
 method = 'push'
 
-[promote-lockfiles]
-source-ref = 'bodhi-updates'
-target-ref = 'testing-devel'
+# We're not using this anymore, likely will delete it eventually
+#[promote-lockfiles]
+#source-ref = 'bodhi-updates'
+#target-ref = 'testing-devel'
+#trigger.mode = 'periodic'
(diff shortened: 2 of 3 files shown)


config-bot/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:30
+FROM quay.io/coreos-assembler/coreos-assembler:master
```
dustymabe: Hmm. Things just got a lot more heavyweight :) - Does this mean we also require /dev/kvm for config-bot now?

dustymabe:
Could we just make a separate pipeline that uses bodhi-updates as the source and just runs `fetch --update-lockfiles` and then config bot picks up that output?

IOW we keep promote-lockfiles and run the `fetch --update-lockfiles` somewhere else rather than as part of config bot

config-bot/Dockerfile
jlebon: I originally wanted to use cosa for this, and had a patch to drop the req. on `/dev/kvm` for our needs. :) But in the end, I think it's simpler to just invoke `rpm-ostree` ourselves directly.

jlebon:
> Could we just make a separate pipeline that uses bodhi-updates as the source and just runs `fetch --update-lockfiles` and then config bot picks up that output?

That could work, though hmm... feels like overkill if we can just run that one `rpm-ostree` command here.

jlebon:
I think an MVP for this is just doing the PR workflow so we're still covered by CI. Do a dumb detection if a PR is already opened, and just update that one if so.

We can then refine it later on to do the "branch push and silently merge if CI passes and it's trivial package bumps, otherwise open a PR".

dustymabe:
Sounds good

jlebon:
Going to pick this up again soon!

jlebon:
Replaced by https://github.com/coreos/fedora-coreos-pipeline/pull/229.

PR #50, opened 2019-10-21
scripts: add versionary.py to drive FCOS versioning

jlebon:
This script implements the versioning scheme as described in:
coreos/fedora-coreos-tracker#211

The next step is to use this in the FCOS pipeline to drive versioning.

### DIFF
--- a/scripts/versionary.py
+++ b/scripts/versionary.py
@@ -1,0 +1,175 @@
+#!/usr/bin/python3 -u
+
+'''
+    Implements the Fedora CoreOS versioning scheme as per:
+        https://github.com/coreos/fedora-coreos-tracker/issues/81
+        https://github.com/coreos/fedora-coreos-tracker/issues/211
+'''
+
+import argparse
+import json
+import os
+import re
+import subprocess
+import sys
+import time
+import yaml
+
+from datetime import datetime
+
+# streams which don't use lockfiles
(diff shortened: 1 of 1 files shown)


jlebon:
Marking as draft for now since the strategy used for the Y component is still being discussed. @bgilbert had a proposal of instead making the timestamp an explicit separate file instead which sits alongside the lockfile.

dustymabe:
> @bgilbert had a proposal of instead making the timestamp an explicit separate file instead which sits alongside the lockfile.

This seems like the option with the least "magic".

jlebon:
Hmm, we could also just make rpm-ostree inject a `timestamp` key in the lockfile itself and base versioning off of that. Probably something it should do anyway.

I'd like to avoid two separate files that can go out of sync if possible.

jlebon:
@bgilbert Thoughts on https://github.com/coreos/fedora-coreos-releng-automation/pull/50#issuecomment-549423766 ?

bgilbert:
Works for me.

jlebon:
OK, reworked this! Now requires https://github.com/coreos/rpm-ostree/pull/1938.
Will lift WIP once I've tested it some more.

jlebon:
OK, this is ready to go!

jlebon:
Thanks for the review!

Issue #51, opened 2019-10-22
coreos-koji-tagger: perform an initial run on startup


dustymabe:
Let's perform an initial run where we check the current lockfiles in the repo and tag any packages that need it. This would be nice in case the tagger my have missed some events or been down for a period of time.

PR #55, opened 2019-11-28
coreos-koji-tagger: process lockfiles on startup

jlebon:
In the event that the app goes down or misses a message, it's important
that we sync up with the latest state on restart. Otherwise, we might be
unsynced until the next fedmsg comes along.

Closes: #51


jlebon:
Tested working!

PR #58, opened 2019-12-06
coreos-koji-tagger: move to f31 and add catch_exceptions_and_continue decorator

dustymabe:
- coreos-koji-tagger: add a catch_exceptions_and_continue decorator
- coreos-koji-tagger: logging.warn is deprecated
- coreos-koji-tagger: switch to fedora 31


jlebon:
(Though the discussions in https://github.com/coreos/fedora-coreos-releng-automation/pull/45#discussion_r328297631 is relevant here too re. that last commit.)

PR #59, opened 2019-12-06
coreos-koji-tagger: support yaml lockfiles

dustymabe:
- coreos-koji-tagger: update instructions for testing in stage
- coreos-koji-tagger: rework test setup using Mock
- coreos-koji-tagger: support yaml lockfiles

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -570,9 +570,7 @@ if __name__ == '__main__':
     # Mock the web request to get the data so that we can easily
     # modify the below values in order to run a test:
     from unittest.mock import Mock
-
-    requests_response = Mock()
-    requests_response.text = """
+    sample_lockfile = """
 packages:
   GeoIP:
     evra: 1.6.12-5.fc30.x86_64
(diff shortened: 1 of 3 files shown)


dustymabe:
this builds on #58 - please see individual commit messages and review only the last 3 commits

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -305,15 +305,18 @@ def process_lockfiles(self, commit=None):
...
+                    logger.info(f'Attempting to retrieve data from {url}')
+                    r = requests.get(url)
+                    if r.ok:
+                        # parse the lockfile and add the set of rpm NEVRAs (strings)
+                        desiredrpms.update(parse_lockfile_data(r.text, filetype))
+                        break # If both yaml and json files exist, only parse one 
+                              # of them. Prefer yaml.
+                    else:
+                        # Log any errors we encounter. 404s are ok, but won't hurt to log
+                        logger.warning('URL request error: %s' % r.text.strip())
```
jlebon: Not directly related, though can we do a check on whether `desiredrpms` is empty here and return early if so?

dustymabe: IIUC the only reason that `desiredrpms` would be empty is if no lockfiles existed in the repo at all. Does that match your understanding?

If so, is targeting a repo with no lockfiles valid. If valid is it something we want to optimize for?

jlebon: I was thinking more about it in terms of reasoning through the code than optimizing for it. Getting the trivial cases out of the way means we don't have to think about them when hacking on code lower down.

dustymabe: ok. I'll address in a followup: https://github.com/coreos/fedora-coreos-releng-automation/pull/62

dustymabe: Sound good?

PR #60, opened 2019-12-06
coreos-koji-tagger: call distrepo manually

dustymabe:
- coreos-koji-tagger: timeout waiting for signing after 10 minutes
- coreos-koji-tagger: call distrepo manually

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -398,16 +398,23 @@ class Consumer(object):
                 # races in tag2distrepo. https://pagure.io/koji/issue/1630
                 # Before running distrepo let's wait for all rpms to
                 # pass through signing and make it into the target tag
-                while True:
+                #
+                # If not done in ten minutes then just timeout (60*10s = 10 minutes)
+                for x in range(0, 60):
                     currentbuildids = self.get_tagged_buildids(self.target_tag)
                     difference = desiredbuildids - currentbuildids
                     if difference:
                         logger.info('Waiting on builds to be signed')
                         logger.info('Remaining builds: %s' %
                                         [buildsinfo[x].nvr for x in difference])
-                        time.sleep(5)
+                        time.sleep(10)
                         continue
                     break
+                # If all the builds didn't make it into the target
+                # then just return here.
+                if difference:
+                    logger.error('Some builds never got signed..  Giving up')
+                    return
                 # This code is mostly stolen from:
                 # https://pagure.io/releng/tag2distrepo/blob/master/f/tag2distrepo.py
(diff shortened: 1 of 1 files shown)


dustymabe:
this builds on #59 - please see individual commit messages and review only the last 2 commits

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -393,6 +394,41 @@ def process_lockfiles(self, commit=None):
...
+                    'inherit': False,
+                    'latest': False,
+                    'multilib': False,
+                    'split_debuginfo': False,
+                    'skip_missing_signatures': False,
+                    'allow_missing_signatures': False,
+                }
+                task = self.koji_client.distRepo(self.target_tag,
+                                                    keys, **task_opts)
+                watch_tasks(self.koji_client, [task], poll_interval=10)
```
jlebon: One improvement here is to split out waiting for the tags and doing `distRepo` on a separate thread so that the processing time for each (relevant) message stays low. Otherwise, there's a theoretical issue with messages queuing up faster than we're clearing them. I don't think we'll hit this in practice today since we don't have super high churn on the lockfiles.

dustymabe: I think right now we want to keep it so that requests are processed serially so that more than one distrepo can't be running at the same time. Do you agree?

Keeping message processing time low would be nice but I think your proposal could introduce a race condition again, right?

jlebon: Right yeah, it'd require careful implementation so we don't have a race. What I was thinking was a shared list object where the tagger just appends the rpms to wait for, and the thread would just take out of the list, wait for them to enter the pool, then run `distRepo`. (And e.g. the tagger could add more rpms on the list while the waiting or `distRepo` is happening for the next iteration on the thread main loop).

jlebon: To be clear, I think it'd be cool, but I definitely don't think it's worth blocking this!

dustymabe: Yeah that seems interesting, but probably better to keep this simple especially since we don't have to worry about lost messages because we're sitting on top of AMQP already. We can let AMQP handle the complexity. The only real benefit I see to adding complexity here would be to improve the performance of the koji tagger. Right now our bodhi run (the input to the lockfiles getting changed) takes a long time (much longer than the koji tagger) so I'd say it's a problem we won't really see until we start updating more and more lockfiles.

dustymabe:
ok I think this is approved and all outstanding code review comments have been discussed/addressed. I will merge this after #59 merges

Issue #46, opened 2019-10-02
coreos-koji-tagger: handle koji restarts gracefully


jlebon:
Right now if Koji restarts, we just lose auth. We should detect this somehow and e.g. re-auth as needed.

dustymabe:
Yep. This happened to us recently when koji was updated/restarted: https://pagure.io/fedora-infrastructure/issue/8244

dustymabe:
until we fix it (soon!) you can get around this in two ways:

- start a new build and the new build will trigger a new rollout (easy)
- delete project and re-create it (big hammer)
   - you'll need [2FA set up](https://fedoraproject.org/wiki/Infrastructure_Two_Factor_Auth) 
   - log in to `batcave01.phx2.fedoraproject.org` (make sure you have [ssh-config set up](https://fedora-infra-docs.readthedocs.io/en/latest/sysadmin-guide/sops/sshaccess.html))
   - execute `sudo rbac-playbook -l os_masters[0] -t delete openshift-apps/coreos-koji-tagger.yml`       
   - execute `sudo rbac-playbook -l os_masters[0] openshift-apps/coreos-koji-tagger.yml`

bgilbert:
@dustymabe After the new rollout, will coreos-koji-tagger automatically tag the packages that it missed?

dustymabe:
> @dustymabe After the new rollout, will coreos-koji-tagger automatically tag the packages that it missed?

Not currently, but I have thought it would be a good idea to do.. Opened an RFE for it: https://github.com/coreos/fedora-coreos-releng-automation/issues/51

jlebon:
See follow-up issue to this in https://github.com/coreos/fedora-coreos-releng-automation/issues/70.

PR #61, opened 2019-12-06
coreos-koji-tagger: re-auth if koji get's restarted

dustymabe:
The coreos-koji-tagger is a long running process. If koji gets restarted
then we end up with a stale session and we won't make any progress.
Let's try to re-auth if we ever get a denial.

Fixes: https://github.com/coreos/fedora-coreos-releng-automation/issues/46

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -251,11 +251,11 @@ class Consumer(object):
         # If a keytab was specified let's try to auth.
         self.keytab_file = os.getenv('COREOS_KOJI_TAGGER_KEYTAB_FILE')
         if self.keytab_file:
+            # Assert the defined keytab file exists
             if not os.path.exists(self.keytab_file):
                 raise Exception("The specified keytab file "
                                 "does not exist: %s" % self.keytab_file)
-            principal = find_principal_from_keytab(self.keytab_file)
-            self.koji_client.gssapi_login(principal, self.keytab_file)
+            self.koji_login()
         else:
             logger.info('No keytab file defined in '
                         '$COREOS_KOJI_TAGGER_KEYTAB_FILE')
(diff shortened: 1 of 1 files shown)


dustymabe:
this builds on #60 - please see individual commit messages and review only the last commit

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -436,6 +434,20 @@ def process_lockfiles(self, commit=None):
...
 
+    def koji_login(self):
+        # If no keytab then just quietly return
+        if not self.keytab_file:
+            return
+        # If already authenticated then nothing to do
+        if self.koji_client.getLoggedInUser():
+            return
+        # Assert the defined keytab file exists
+        if not os.path.exists(self.keytab_file):
```
jlebon: Hmm, let's check this up front at `__init__` time?

dustymabe: we do check it at init time AND here. At init time we check it and print out a useful message to the logs. Here we just silently ignore it. Basically running with no keytab wile is a valid case.

One alternative is that we conditionalize the call to `koji_login()` to only call it if the keytab file var is populated.

```
if self.keytab_file:
    self.koji_login()
```

jlebon: What I mean is checking that the keytab file itself exists. Right now, we're checking this on every re-login instead of once on init. Which is fine, just interesting I guess, since it implies we're guarding ourselves from the keytab being deleted while we're running live, which is "above and beyond the call of duty" for this kind of code. :)

dustymabe: But we're not checking if the keytab file exists. We're checking if the `self.keytab_file` variable is not empty (lower cost than checking if the file exists). I don't know of a better way to do the conditional logic.

dustymabe: wow.. I totally mistook your original comment.

dustymabe: ok addressed now. Sorry for misunderstanding the original comment

dustymabe:
I need to re-run this through testing

dustymabe:
> I need to re-run this through testing

seems to be ok

PR #62, opened 2019-12-09
coreos-koji-tagger: return early if no desiredrpms detected

dustymabe:
This shouldn't happen, but we'll return early if we couldn't detect
any desired rpms from the lockfiles in the repo.

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -322,6 +322,12 @@ class Consumer(object):
                     else:
                         # Log any errors we encounter. 404s are ok, but won't hurt to log
                         logger.warning('URL request error: %s' % r.text.strip())
+        if not desiredrpms:
+            logger.warning('No locked RPMs found!')
+            logger.warning("Does the repo:ref (%s:%s) have any lockfiles?" %
+                            (self.github_repo_fullname, self.github_repo_branch))
+            logger.warning('Continuing...')
+            return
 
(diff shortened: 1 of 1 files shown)


dustymabe:
I still need to test this.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -316,6 +316,12 @@ def process_lockfiles(self, commit=None):
                     else:
                         # Log any errors we encounter. 404s are ok, but won't hurt to log
                         logger.warning('URL request error: %s' % r.text.strip())
+        if not desiredrpms:
+            logger.warning('We detected no desiredrpms.')
```
jlebon: Or maybe just "No locked RPMs found." ?

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -316,6 +316,12 @@ def process_lockfiles(self, commit=None):
...
+            logger.warning("Does the repo/ref: %s/%s have any lockfiles?" %
+                            (self.github_repo_branch, self.github_repo_fullname))
```
jlebon: Those args look swapped? (Also, why not use f-strings? I find them much nicer to write/read!)

dustymabe: Sorry copy/paste error.

I like f strings too, but sometimes the lines get too long and breaking them up sucks too.

Issue #32, opened 2019-08-19
Use PR workflow for updating the lockfile


jlebon:
Right now, `config-bot` only supports directly promoting the lockfile using `git push`. We should enhance it so that it instead opens up a PR to bump the lockfile. This unlocks (pun intended) a few things:
- a better CI workflow: we can actually test the OS against the latest base content before absorbing it
- smarter integration with coreos-koji-tagger: right now, it tags packages once the lockfile is merged, at which point builds won't actually succeed until the packages are signed (if not already), moved to the coreos-pool tag, and the coreos-pool repo is respinned. This means that there's a window of time during which `testing-devel` would be unbuildable. Switching to a PR workflow would allow e.g. a merge bot to do the tagging and waiting for the repo respin *before* doing the merge, so that `testing-devel` can always be built.

jlebon:
Just mentioning this here for lack of a better place. Part of working on this, we should look at how we expect PR testing to work such that we can still get meaningful CI testing *before* merging. (See e.g. https://github.com/coreos/fedora-coreos-config/pull/186#issuecomment-535572028).

jlebon:
> Part of working on this, we should look at how we expect PR testing to work such that we can still get meaningful CI testing before merging.

I think one easy way to do this without having the tagger tag things in before merging is to have CI just go and fetch all the RPMs mentioned in the overrides directly from Koji and drop them in `overrides/rpm`.

dustymabe:
> I think one easy way to do this without having the tagger tag things in before merging is to have CI just go and fetch all the RPMs mentioned in the overrides directly from Koji and drop them in `overrides/rpm`.

that's a good idea!

jlebon:
This is obsolete now that we've moved lockfile bumping to https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile.

PR #63, opened 2019-12-09
scripts: add download-overrides.py

jlebon:
I intend to use this to close the gap we have in upstream CI testing
when adding overrides:

https://github.com/coreos/fedora-coreos-releng-automation/issues/32#issuecomment-541134070

### DIFF
--- a/scripts/download-overrides.py
+++ b/scripts/download-overrides.py
@@ -1,0 +1,53 @@
+#!/usr/bin/python3
+
+import dnf.subject
+import hawkey
+import os
+import yaml
+import subprocess
+import sys
+
+
+# this was copied from coreos-koji-tagger
+def get_NVRA_from_NEVRA(string: str) -> str:
+    form = hawkey.FORM_NEVRA
+
+    # get a hawkey.Subject object for the string
+    subject = dnf.subject.Subject(string)  # returns hawkey.Subject
+
+    # get a list of hawkey.NEVRA objects that are the possibilities
+    nevras = subject.get_nevra_possibilities(forms=form)
+
+    # return the first hawkey.NEVRA item in the list of possibilities
+    rpminfo = nevras[0]
+
+    # come up with rpm NVRA
+    nvra = f"{rpminfo.name}-{rpminfo.version}-{rpminfo.release}.{rpminfo.arch}"
+    return nvra
+
+
+def is_override_lockfile(filename: str) -> bool:
+    return (filename.startswith('manifest-lock.overrides.')
+            and filename[-4:] in ['json', 'yaml'])
+
+
+assert os.path.isdir("builds"), "Missing builds/ dir; is this a cosa workdir?"
+
+rpms = set()
+for filename in os.listdir(os.path.join("src/config")):
+    if is_override_lockfile(filename):
+        with open(f'src/config/{filename}') as f:
+            lockfile = yaml.safe_load(f)
+        for pkg, pkgobj in lockfile['packages'].items():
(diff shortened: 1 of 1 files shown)


scripts/download-overrides.py
```diff
@@ -0,0 +1,56 @@
...
+    form = hawkey.FORM_NEVRA
+
+    # get a hawkey.Subject object for the string
+    subject = dnf.subject.Subject(string)  # returns hawkey.Subject
+
+    # get a list of hawkey.NEVRA objects that are the possibilities
+    nevras = subject.get_nevra_possibilities(forms=form)
+
+    # get a list of hawkey.NEVRA objects that are the possibilities
+    nevras = subject.get_nevra_possibilities(forms=form)
```
dustymabe: https://github.com/coreos/fedora-coreos-releng-automation/pull/64

scripts/download-overrides.py
```diff
@@ -0,0 +1,56 @@
...
+if not rpms:
+    print("No overrides; exiting.")
+    sys.exit(0)
+
+# could probably be more efficient here by using the Koji API directly, but
+# meh... there shouldn't be that many overrides anyway
+for rpm in rpms:
+    os.makedirs('overrides/rpm', exist_ok=True)
+    subprocess.check_call(['koji', 'download-build', '--rpm', rpm],
+                          cwd='overrides/rpm')
```
dustymabe: For builds that aren't noarch I think you want to specify `--arch`. Otherwise you get a lot of "extra".

scripts/download-overrides.py
jlebon: Ahh heh, updated!

scripts/download-overrides.py
jlebon: `download-build --rpm` already only takes a full NEVRA, so it only downloads that specific RPM:

```
$ koji download-build --rpm ignition-2.0.1-8.gita8f91fa.fc31.x86_64
Downloading: ignition-2.0.1-8.gita8f91fa.fc31.x86_64.rpm
[====================================] 100%   3.71 MiB
$ ls
ignition-2.0.1-8.gita8f91fa.fc31.x86_64.rpm
```

dustymabe: ahh - I was mistaking this with `koji download-build` without the `--rpm`. Thanks for the clarification.

scripts/download-overrides.py
```diff
@@ -0,0 +1,53 @@
...
+            and filename[-4:] in ['json', 'yaml'])
+
+
+assert os.path.isdir("builds"), "Missing builds/ dir; is this a cosa workdir?"
+
+rpms = set()
+for filename in os.listdir(os.path.join("src/config")):
+    if is_override_lockfile(filename):
+        with open(f'src/config/{filename}') as f:
+            lockfile = yaml.safe_load(f)
```
dustymabe: just thought of this new piece:

on lins 31 above we support yaml or json, but here we are just doing `yaml.safe_load(f)`. We should handle both cases?

jlebon: Heh, here we're relying on the fact that JSON is a subset of YAML:

```
In [1]: import yaml

In [2]: with open('manifest-lock.overrides.x86_64.json') as f:
   ...:     j = yaml.safe_load(f)
   ...:

In [3]: j
Out[3]: {'packages': {'ignition': {'evra': '2.0.1-8.gita8f91fa.fc31.x86_64'}}}
```

dustymabe: wow didn't realize that worked. Nice

PR #65, opened 2019-12-16
coreos-koji-tagger: catch koji.AuthError exception

dustymabe:
In our logic to determine if we're still logged in we need to catch
the koji.AuthError exception and pass so that we can fall through to
the code that does the login.

This is a followup to d46a3a1 which was part of
https://github.com/coreos/fedora-coreos-releng-automation/issues/46


dustymabe:
doing a bit of testing before merge.

PR #66, opened 2019-12-18
coreos-ostree-importer: switch to plain http(s) url

dustymabe:
We'll send the messages with a plain http(s) url so that this importer
doesn't need to be configured with AWS credentials.

### DIFF
--- a/coreos-ostree-importer/coreos_ostree_importer.py
+++ b/coreos-ostree-importer/coreos_ostree_importer.py
@@ -1,7 +1,5 @@
 #!/usr/bin/python3
 
-import boto3
-import botocore
 import fedora_messaging
 import fedora_messaging.api
 import hashlib
@@ -13,6 +11,7 @@ import sys
 import tarfile
 import tempfile
 import traceback
+import urllib.request
 
 # Set local logging
 logger = logging.getLogger(__name__)
@@ -33,7 +32,7 @@ EXAMPLE_MESSAGE_BODY = json.loads("""
     "build_id": "30.20190905.0",
     "stream": "testing",
     "basearch": "x86_64",
-    "commit": "s3://fcos-builds/prod/streams/testing/builds/30.20190905.0/x86_64/ostree-commit.tar",
+    "commit": "https://fcos-builds/prod/streams/testing/builds/30.20190905.0/x86_64/ostree-commit.tar",
     "checksum": "sha256:d01db6939e7387afa2492ac8e2591c53697fc21cf16785585f7f1ac0de692863",
     "ostree_ref": "fedora/x86_64/coreos/testing",
     "ostree_checksum": "b4beca154dab3696fd04f32ddab818102caa9247ec3192403adb9aaecc991bd9",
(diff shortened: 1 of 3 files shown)


dustymabe:
Also, switch to Fedora 31

lucab:
@dustymabe can we use https://builds.coreos.fedoraproject.org/ as the base URL perhaps?

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -184,23 +183,13 @@ def get_sha256sum(filepath: str) -> str:
...
     filepath = os.path.join(tmpdir, filename)
 
-    # Grab file from s3
-    logger.info(f"Downloading object from s3: {url}")
-    s3 = boto3.client("s3")
-    bucket, key = parse_s3_url(url)
-    s3.download_file(bucket, key, filepath)
+    # Grab file from the url
+    logger.info(f"Downloading object from url: {url}")
+    urllib.request.urlretrieve(url, filepath)
```
jlebon: Just a side note, but we should also download the signature and verify it before importing. E.g. we can add a `commit-sig` to the message?

dustymabe:
> @dustymabe can we use https://builds.coreos.fedoraproject.org/ as the base URL perhaps?

yep, added a commit on top that updates the example fedora message.

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: We already check the sha256sum. The only thing I can think that the signature file would help us with is to make sure only signed content got into the repo (i.e. if we're worried about an attacker trying to send a message to import invalid content), but I can think of even better ways to do that too. One of them would be to verify the commit in the ostree repo after the file was untarred, but before it was imported.

jlebon: > One of them would be to verify the commit in the ostree repo after the file was untarred, but before it was imported.

Hmm, I thought that was already the case. We definitely should do this too if not. Ahh OK yup, we need to explicitly do `--gpgverify`.

> We already check the sha256sum. The only thing I can think that the signature file would help us with is to make sure only signed content got into the repo (i.e. if we're worried about an attacker trying to send a message to import invalid content)

One argument for requiring a signature on the tarball itself is to sanity check that [it's not malicious](https://en.wikipedia.org/wiki/Tar_(computing)#Tarbomb) before trying to extract it. Actually, even the docs for [TarFile.extractall()](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall) has a warning related to this.

So I think we should do both.

dustymabe: so:

- remove checksum and apply signature checking for the tarball
- add commit checking in extracted ostree repo

?

dustymabe: pushing that bit to a followup: https://github.com/coreos/fedora-coreos-releng-automation/issues/68

PR #67, opened 2019-12-19
versionary: use `-` to separate build iteration for semver

jlebon:
This changes the versioning scheme to use `-` instead of `.` for the
final element so that the version string is a valid semver string (with
the understanding that our OS *isn't* bound to semver constraints,
though it's useful for e.g. parsers and higher-level code like OKD's CI
which expect all components to respect it).

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/338


jlebon:
Let's hold this until there's more consensus on that ticket.

jlebon:
We're not doing this anymore as per discussions in ticket.

PR #74, opened 2020-02-18
Dockerfiles: remove the `dnf clean all` where it doesn't make sense

dustymabe:
Rather than hit our fedora servers 2 or 3 times for the metadata let's
just let the cache bake in the images.

### DIFF
--- a/coreos-koji-tagger/Dockerfile
+++ b/coreos-koji-tagger/Dockerfile
@@ -5,19 +5,19 @@ FROM registry.fedoraproject.org/fedora:31
 ENV PYTHONUNBUFFERED=true
 
 # Get any latest updates since last container spin
-RUN dnf update -y && dnf clean all
+RUN dnf update -y
 
 # Install necessary (fedmsg/koji/yaml) libraries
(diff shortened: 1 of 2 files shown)


jlebon:
How about we just combine the `dnf` invocations into a single `RUN`, which has `dnf clean all` at the end? We don't need it in the final image, and doing it all in one shot means we keep it out of the layers too.

dustymabe:
> How about we just combine the `dnf` invocations into a single `RUN`, which has `dnf clean all` at the end? We don't need it in the final image, and doing it all in one shot means we keep it out of the layers too.

Keeping the cache in the layers is not as big of a deal to me because because these images, once built, stay local to the cluster. For that reason, I'd prefer to keep the lines separate because they each have separate comments and it's ugly when you combine them. I wish Dockerfiles were a little more scriptable and you could just define the rpms in an array at the top of the file with comments interspersed. Buildah can do this but it's not quite as popular and not supported in openshift.

jlebon:
We could work around that by using a separate file for specifying deps, like cosa has. But yeah, we don't need to block on that.

PR #76, opened 2020-02-18
coreos-ostree-importer: sender: switch to waiting on request response

dustymabe:
A few commits. The most important being the switch to waiting on a response
that the ostree importer has finished before continuing.

- coreos-ostree-importer: sender: bump timeout to 10 minutes
- coreos-ostree-importer: sender: switch to waiting on request response
- coreos-ostree-importer: sender: clean up unrelated comment

### DIFF
--- a/coreos-ostree-importer/send-ostree-import-request.py
+++ b/coreos-ostree-importer/send-ostree-import-request.py
@@ -14,7 +14,7 @@ import sys
 # Pick up libraries we use that are delivered along with COSA
 sys.path.insert(0, '/usr/lib/coreos-assembler')
 from cosalib.meta import GenericBuildMeta
-from cosalib.fedora_messaging_request import send_message
+from cosalib.fedora_messaging_request import send_request_and_wait_for_response
 
 # Example datagrepper URLs to inspect sent messages:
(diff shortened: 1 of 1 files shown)


coreos-ostree-importer/send-ostree-import-request.py
```diff
@@ -90,7 +90,6 @@ def get_bucket_and_prefix(path):
 
 def validate_response(response):
     if response["status"].lower() == "failure":
-        # https://pagure.io/robosignatory/pull-request/38
         if "failure-message" not in response:
             raise Exception("Signing failed")
         raise Exception(f"Signing failed: {response['failure-message']}")
```
jlebon: Seems like these should be updated too to something like `Importing failed` ?

dustymabe: +1 - added a new commit with the fix

dustymabe:
I'm leaving this as a draft because we don't want to merge it until the prod repos have been migrated and we are running coreos-ostree-importer in prod. Otherwise our pipeline will get stuck because no one will be there to reply to the request.

dustymabe:
This was implemented in #97 #100 and #101

PR #77, opened 2020-02-19
coreos-ostree-importer: configure/check for group writable 

dustymabe:
We are going to be running two applications with access to the ostree repos:
`coreos-ostree-importer` and `fedora-ostree-pruner`. In order to do this in
OpenShift we need everything to be group writable. This PR sets the umask
such that new files are creating group writable and also adds a check to make
sure that directories in the repo have the group writable bit set.

### DIFF
--- a/coreos-ostree-importer/coreos_ostree_importer.py
+++ b/coreos-ostree-importer/coreos_ostree_importer.py
@@ -6,6 +6,7 @@ import hashlib
 import json
 import logging
 import os
+import stat
 import subprocess
 import sys
 import tarfile
@@ -48,13 +49,19 @@ KNOWN_OSTREE_REPOS = {
     "compose": "/mnt/koji/compose/ostree/repo",
 }
 
-
 class Consumer(object):
     def __init__(self):
-        # Check the possible repos to make sure they exist
+        # Do sanity checks on the repos
         for path in KNOWN_OSTREE_REPOS.values():
+
+            # Check the repo to make sure it exists
             if not ostree_repo_exists(path):
                 raise Exception(f"OSTree repo does not exist at {path}")
+
+            # Sanity check the repo to make sure all directories in the repo
+            # have the appropriate permissions (most importantly group writable).
+            # See https://pagure.io/releng/issue/8811#comment-616490
+            assert_dirs_permissions(path)
 
         logger.info(
             "Processing messages with topic: %s" % FEDORA_MESSAGING_TOPIC_LISTEN
@@ -102,6 +109,11 @@ class Consumer(object):
         target_repo_path = KNOWN_OSTREE_REPOS[target_repo]
         source_repo_path = None
 
+        # Sanity check the repo to make sure all directories in the repo
+        # have the appropriate permissions (most importantly group writable).
+        # See https://pagure.io/releng/issue/8811#comment-616490
+        assert_dirs_permissions(target_repo_path)
+
         logger.info(
             f"Processing request to import {build_id} into the "
             f"{ostree_ref} branch of the {target_repo} repo."
@@ -146,6 +158,18 @@ class Consumer(object):
                 srcrepo=source_repo_path,
                 branch=ostree_ref,
             )
+
+
+def assert_dirs_permissions(path: str):
+    founderror = False
+    for root, dirs, files in os.walk(path):
+        statinfo = os.stat(root)
+        # Verifies group permissions are 0bXXX111XXX (---rwx---)
+        if ((statinfo.st_mode & stat.S_IRWXG) != stat.S_IRWXG):
+            logger.warning(f"Directory {root} does not have rwx group permissions!")
+            founderror = True
+    if founderror:
+        raise Exception(f"Found directories that did not have rwx group permissions")
 
 
 def runcmd(cmd: list, **kwargs: int) -> subprocess.CompletedProcess:
--- a/coreos-ostree-importer/Dockerfile
+++ b/coreos-ostree-importer/Dockerfile
@@ -7,8 +7,12 @@ ENV PYTHONUNBUFFERED=true
 # Get any latest updates since last container spin
 RUN dnf update -y
 
-# Install boto/fedmsg/ostree libraries
-RUN dnf -y install fedora-messaging ostree
+# Install needed rpms
+# Grab ostree from koji for now for https://github.com/ostreedev/ostree/pull/1984
+RUN dnf -y install \
+        fedora-messaging \
+        https://kojipkgs.fedoraproject.org//packages/ostree/2020.1/2.fc31/x86_64/ostree-2020.1-2.fc31.x86_64.rpm \
+        https://kojipkgs.fedoraproject.org//packages/ostree/2020.1/2.fc31/x86_64/ostree-libs-2020.1-2.fc31.x86_64.rpm
 
 # Configure a umask of 0002 which will allow for the group permissions
 # to include write for newly created files. We need this because we'd
--- a/coreos-ostree-importer/Dockerfile
+++ b/coreos-ostree-importer/Dockerfile
@@ -10,6 +10,18 @@ RUN dnf update -y
 # Install boto/fedmsg/ostree libraries
 RUN dnf -y install fedora-messaging ostree
 
+# Configure a umask of 0002 which will allow for the group permissions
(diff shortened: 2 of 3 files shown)


dustymabe:
ok this looks good in testing

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -48,14 +49,34 @@
...
+        # Do sanity checks on the repos
+        for repo, path in KNOWN_OSTREE_REPOS.items():
+
+            # Check the repo to make sure it exists
             if not ostree_repo_exists(path):
                 raise Exception(f"OSTree repo does not exist at {path}")
 
+            # Sanity check the repo to make sure all directories in the repo
+            # are group writable. https://pagure.io/releng/issue/8811#comment-616490
+            # Wrap in try/except so we can hopefull see the logs before openshift
```
jlebon: hopefully

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -48,14 +49,34 @@
...
+            # deletes the pod
+            try:
+                assert_dirs_group_writable(repo)
+            except Exception as e:
+                logger.error(f"Exception detected: {str(e)}")
+                logger.info(f"Sleeping for 30s")
+                time.sleep(30)
+                raise(e)
+
+        # Configure a umask of 0002 which will allow for the group permissions
```
jlebon: Hmm, feels odd to do this from a class which gets run as part of another app. This is normally done early on in an app startup to avoid issues with multi-threading. A safer approach is to make this part of a tiny init script that does `umask` and then `exec`s `fedora-messaging`. Or could just do it in the Dockerfile too, e.g.:

```
CMD ["sh", "-c", "umask 0002 && exec fedora-messaging consume --callback=coreos_ostree_importer:Consumer"]
```

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -148,6 +173,18 @@ def process(self, message: fedora_messaging.api.Message):
             )
 
 
+def assert_dirs_group_writable(repo: str):
```
jlebon: Minor: IMO it would be cleaner if this function just took the path to the repo to check to make it more self-contained. We already have it handy in both callsites.

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -148,6 +173,18 @@ def process(self, message: fedora_messaging.api.Message):
...
+    founderror = False
+    for root, dirs, files in os.walk(KNOWN_OSTREE_REPOS[repo]):
+        statinfo = os.stat(root)
+        if ((statinfo.st_mode & int('0b000000000010000', 2)) == 0):
```
jlebon: We should probably also check `rx` too here. Could just use [`stat.S_IRWXG`](https://docs.python.org/3.8/library/stat.html#stat.S_IRWXG).

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -48,14 +49,34 @@
...
+        for repo, path in KNOWN_OSTREE_REPOS.items():
+
+            # Check the repo to make sure it exists
             if not ostree_repo_exists(path):
                 raise Exception(f"OSTree repo does not exist at {path}")
 
+            # Sanity check the repo to make sure all directories in the repo
+            # are group writable. https://pagure.io/releng/issue/8811#comment-616490
+            # Wrap in try/except so we can hopefull see the logs before openshift
+            # deletes the pod
```
jlebon: Hmm, but OpenShift doesn't delete the pod on failure. It just restarts it. In this case, since we're re-raising the issue, it'll just restart it a bunch of time before failing, but the pod should stick around for us to fetch the logs. So I think we don't need to wrap this in a `try-except` block at all, right?

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: Part of this PR configures the container itself (via /etc/profile.d) to have a umask of 0002. Doing it here too was was mostly a precaution in case the app ever got run outside of the container for whatever reason. We can probably just delete this.

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: +1

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: when it starts a new pod it deletes the old one, doesn't it? I don't know, i've just had a ton if issues in the past when trying to get logs and the pods keep disappearing

coreos-ostree-importer/coreos_ostree_importer.py
jlebon: Hmm, I think that's just an artifact of how Docker runs containers that don't use brackets (so it shells out to Bash, which sources the profile). Not sure if it's something we'd actually want to rely on.

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: ok will do

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: ok

coreos-ostree-importer/coreos_ostree_importer.py
jlebon: The default behaviour is to restart the same pod (i.e. not delete and create a new one). If it keeps restarting, it'll eventually stop with `CrashLoopBackOff` and just sit there.

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: I put it in `/etc/profile.d/umask-for-openshift.sh` with a reason (partly so anyone running `oc rsh` will also get `umask 0002`) so we definitely need it there. Right now I also have it in `coreos_ostree_importer.py` (as good measure because the ` /etc/profile.d/umask-for-openshift.sh` is effective enough to apply here too) but you think it shouldn't be there and you'd rather have it as `CMD ["sh", "-c", "umask 0002 && exec fedora-messaging consume --callback=coreos_ostree_importer:Consumer"]` in the Dockerfile itself because we shouldn't rely on the *non `[]` arguments to command being run as a shell* behavior?

coreos-ostree-importer/coreos_ostree_importer.py
dustymabe: ok I'll remove the try/except

coreos-ostree-importer/coreos_ostree_importer.py
jlebon: I think having it in `/etc/profile.d` makes sense as you said to have it baked in the image. But for the app itself, given that the proper `umask` is critical here, it seems safer to have the app do that itself. The problem here though is that we're not actually in control of the app being launched.

Anyway, let's keep the `os.umask` here for now, and then in a follow-up we change the entrypoint to the app be a simple wrapper shell/python script that just sets umask and `exec`s `fedora-messaging`?

dustymabe: added a `coreos-ostree-importer-wrapper` script that should take care of the concerns here

dustymabe:
just pushed up a change that should address all PR comments. I still need to test this.

coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -148,6 +159,18 @@ def process(self, message: fedora_messaging.api.Message):
             )
 
 
+def assert_dirs_permissions(path: str):
+    founderror = False
+    for root, dirs, files in os.walk(path):
+        statinfo = os.stat(root)
+        # Verifies group permissions are 0bXXX111XXX (---rwx---)
+        if ((statinfo.st_mode & stat.S_IRWXG) == stat.S_IRWXG):
```
jlebon: That should be `!=`, no?

dustymabe: yep. thanks! fixed now

dustymabe:
ok this is tested and ready for merge.

PR #78, opened 2020-02-20
coreos-ostree-importer: support manual runs with datagrepper URL inputs

dustymabe:
This can be useful for testing or if we ever need to manually trigger
imports if the importer fails for any reason like:

- there was a failure when importing a previous version and we want to trigger a new import
- the importer was down and not able to respond to a previous request

Should be able to execute like:

./coreos_ostree_importer.py 'https://apps.fedoraproject.org/datagrepper/id?id=2020-32c268dc-36ba-4cef-be6a-f4969a0c83af&is_raw=true&size=extra-large'

### DIFF
--- a/coreos-ostree-importer/Dockerfile
+++ b/coreos-ostree-importer/Dockerfile
@@ -11,6 +11,7 @@ RUN dnf update -y
 # Grab ostree from koji for now for https://github.com/ostreedev/ostree/pull/1984
 RUN dnf -y install \
         fedora-messaging \
+        python-requests  \
         https://kojipkgs.fedoraproject.org//packages/ostree/2020.1/2.fc31/x86_64/ostree-2020.1-2.fc31.x86_64.rpm \
         https://kojipkgs.fedoraproject.org//packages/ostree/2020.1/2.fc31/x86_64/ostree-libs-2020.1-2.fc31.x86_64.rpm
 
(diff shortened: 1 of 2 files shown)


dustymabe:
this builds on #77 so there are a few extra commits in there until that merges.

dustymabe:
ok, ready for review now

dustymabe:
> Neat, mostly LGTM though a few notes:
> 
> * The commit message says `coreos-koji-tagger`, but it's `coreos-ostree-importer`, right?

good catch. fixed!

> 
> * `s/testint/testing/` in commit message

also fixed

> 
> * Hmm, does this imply that running it directly without passing a URL will make it try to import the example payload? I'm guessing it fails at some point into the operation, but still... feels dangerous to just have that exposed like this if we also want to genuinely support direct invocation as something useful. How about e.g. a global `dry_run` or something that does everything up to but not including the `pull-local` part? We can do this in a follow-up too.

Yes. The EXAMPLE_PAYLOAD is mostly useful only useful for local testing, but **CAN** be useful for real production things too. For example, for some reason in staging (and prod actually) providing links to `https://apps.stg.fedoraproject.org/datagrepper/` doesn't work so if I really did want to import something that had already sent a message out then I'd have to copy in the payload into EXAMPLE_PAYLOAD in the script and then run it (did this several times yesterday). 

I'm not really worried about issues with accidentally running and importing EXAMPLE_PAYLOAD by mistake. We checks to make sure we're doing the right thing. i.e. we check to make sure the parent of the commit we're importing is in the repo and the ref we're targeting is pointing to it.

PR #81, opened 2020-02-25
coreos-ostree-importer: only update summary file for prod repo

dustymabe:
Only create/update a summary file if we are targeting the prod
OSTree repo. The compose repo doesn't use a summary file (i.e.
robosig updates the files under ./refs/heads) and it causes problems
if the summary file exists.

### DIFF
--- a/coreos-ostree-importer/coreos_ostree_importer.py
+++ b/coreos-ostree-importer/coreos_ostree_importer.py
@@ -160,6 +160,11 @@ class Consumer(object):
                 branch=ostree_ref,
             )
 
+        # Update the summary file if we're in the prod repo. The compose
+        # repo doesn't use a summary file and creating one causes problems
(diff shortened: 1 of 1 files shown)


coreos-ostree-importer/coreos_ostree_importer.py
```diff
@@ -160,6 +160,11 @@ def process(self, message: fedora_messaging.api.Message):
                 branch=ostree_ref,
             )
 
+        # Update the summary file if we're in the prod repo. The compose
+        # repo doesn't use a summary file and creating one causes problems
```
jlebon: Optional: link to more background on this?

dustymabe: unfortunately I don't really have one :( I guess I could try to point to some of the code that needs this. ENOTIME for now

Issue #70, opened 2020-01-09
coreos-koji-tagger: error re-authenticating: `Mutual authentication failed`


dustymabe:
So [now that we properly catch the koji.AuthError exception](https://github.com/coreos/fedora-coreos-releng-automation/pull/65) we have moved to the next failure. We apparently hit a problem when trying to authenticate again.

```
2020-01-09 04:03:50,216 INFO coreos_koji_tagger - Received koji.AuthError from koji. Re-attempting login.
2020-01-09 04:03:50,217 INFO coreos_koji_tagger - Running command: ['/usr/bin/klist', '-k', '/etc/coreos-koji-tagger-keytab/koji-keytab']
2020-01-09 04:03:50,568 INFO coreos_koji_tagger - Found principal coreos-koji-tagger/coreos-koji-tagger.fedoraproject.org@FEDORAPROJECT.ORG in keytab
2020-01-09 04:03:50,640 ERROR requests_kerberos.kerberos_ - handle_other(): Mutual authentication failed
2020-01-09 04:03:50,643 ERROR coreos_koji_tagger - Caught Exception!
2020-01-09 04:03:50,643 ERROR coreos_koji_tagger - ###################################
Traceback (most recent call last):
2020-01-09 04:03:50,643 ERROR coreos_koji_tagger - ###################################
2020-01-09 04:03:50,643 ERROR coreos_koji_tagger - 	 continuing...
  File "/usr/lib/python3.7/site-packages/coreos_koji_tagger.py", line 195, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python3.7/site-packages/coreos_koji_tagger.py", line 300, in process_message
    self.koji_login()
  File "/usr/lib/python3.7/site-packages/coreos_koji_tagger.py", line 461, in koji_login
    self.koji_client.gssapi_login(principal, self.keytab_file)
  File "/usr/lib/python3.7/site-packages/koji/__init__.py", line 2415, in gssapi_login
    raise AuthError('unable to obtain a session')
koji.AuthError: unable to obtain a session
```

The interesting message in there is the `requests_kerberos.kerberos_ - handle_other(): Mutual authentication failed` which comes from the [requests-kerberos](https://github.com/requests/requests-kerberos) python library.

I'll dig into this later, for now I'll restart the tagger.

jlebon:
[...]

PR #83, opened 2020-03-02
coreos-koji-tagger: retry koji login for 15m

jlebon:
We're hitting this pretty frequently in prod now. Let's try retrying for
a bit in case it's simply due to Koji restarting or something.

See: https://github.com/coreos/fedora-coreos-releng-automation/issues/70


jlebon:
(Not tested yet)

PR #84, opened 2020-03-10
versionary: stop creating OSTree repo for --print-only

jlebon:
We don't need to do this anymore now:

https://github.com/coreos/rpm-ostree/commit/37a6a3ab696050548578fb15650cbea6540484be


jlebon:
> Can we wait and merge after this next round of builds?

Yup, sure.

PR #86, opened 2020-03-11
scripts: check epoch match for overrides files entry

dustymabe:
This should help us catch issues in the overrides file in CI. See
https://github.com/coreos/fedora-coreos-config/pull/293.

### DIFF
--- a/scripts/download-overrides.py
+++ b/scripts/download-overrides.py
@@ -8,8 +8,8 @@ import subprocess
 import sys
 
 
-# this was copied from coreos-koji-tagger
-def get_NVRA_from_NEVRA(string: str) -> str:
+# this was partially copied from coreos-koji-tagger
+def get_rpminfo(string: str) -> str:
     form = hawkey.FORM_NEVRA
 
     # get a hawkey.Subject object for the string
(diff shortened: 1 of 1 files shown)


scripts/download-overrides.py
```diff
@@ -20,34 +20,46 @@ def get_NVRA_from_NEVRA(string: str) -> str:
...
 for filename in os.listdir(os.path.join("src/config")):
     if is_override_lockfile(filename):
         with open(f'src/config/{filename}') as f:
             lockfile = yaml.safe_load(f)
         for pkg, pkgobj in lockfile['packages'].items():
-            rpms.add(get_NVRA_from_NEVRA(f"{pkg}-{pkgobj['evra']}"))
+            rpminfo = get_rpminfo(f"{pkg}-{pkgobj['evra']}")
+            rpmnvra = f"{rpminfo.name}-{rpminfo.version}-{rpminfo.release}.{rpminfo.arch}"
+            rpms.add(rpmnvra)
+            subprocess.check_call(['koji', 'download-build', '--rpm', rpmnvra], cwd='overrides/rpm')
```
jlebon: Hmm, if we used the Koji API directly we could verify the epoch before having to download the RPM. But meh... we're only continuously doing this in CI right now and not developers' laptops, so we can do that as a follow up!

scripts/download-overrides.py
```diff
@@ -20,34 +20,46 @@ def get_NVRA_from_NEVRA(string: str) -> str:
...
+    return rpminfo
 
 def is_override_lockfile(filename: str) -> bool:
     return (filename.startswith('manifest-lock.overrides.')
             and filename[-4:] in ['json', 'yaml'])
 
+def assert_epochs_match(overrides_epoch: int, rpmfile_epoch: str):
+    # normalize the input into a string
+    if overrides_epoch is None:
+       normalized_overrides_epoch = '(none)'
```
jlebon: Worth mentioning here that this matches the output of `%{E}`.

scripts/download-overrides.py
```diff
@@ -20,34 +20,46 @@ def get_NVRA_from_NEVRA(string: str) -> str:
...
+            rpmnvra = f"{rpminfo.name}-{rpminfo.version}-{rpminfo.release}.{rpminfo.arch}"
+            rpms.add(rpmnvra)
+            subprocess.check_call(['koji', 'download-build', '--rpm', rpmnvra], cwd='overrides/rpm')
+            # Make sure the epoch matches what was in the overrides file
+            # otherwise we can get errors: https://github.com/coreos/fedora-coreos-config/pull/293
+            cp = subprocess.run(['rpm', '-qp', '--queryformat', '%{E}', f'{rpmnvra}.rpm'],
+                                  check=True,
+                                  capture_output=True,
+                                  cwd='overrides/rpm')
+            rpmfile_epoch = cp.stdout.decode('utf-8')
```
jlebon: Optional: could fold into the previous run with e.g. `rpmfile_epoch = subprocess.check_output([...], ..., encoding='utf-8')`.

scripts/download-overrides.py
dustymabe: +1

scripts/download-overrides.py
dustymabe: +1 for follow up

scripts/download-overrides.py
dustymabe: I've tested this like it is so I'll leave it for now

scripts/download-overrides.py
dustymabe: ok I added a comment

PR #89, opened 2020-03-20
coreos-koji-tagger: handle both testing-devel and next-devel

jlebon:
Generalize things a bit so that we can monitor multiple branches, then
add next-devel to the list.

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -45,7 +45,7 @@ COREOS_KOJI_USER = 'coreosbot'
 
 # XXX: should be in config file
 DEFAULT_GITHUB_REPO_FULLNAME = 'coreos/fedora-coreos-config'
-DEFAULT_GITHUB_REPO_BRANCH   = 'refs/heads/testing-devel'
+DEFAULT_GITHUB_REPO_BRANCHES = 'refs/heads/testing-devel refs/heads/next-devel'
 
 # We are processing the org.fedoraproject.prod.github.push topic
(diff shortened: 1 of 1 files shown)


jlebon:
(Not tested yet.)

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -303,10 +304,7 @@ def process_message(self, message: fedora_messaging.api.Message):
         self.process_lockfiles(commit)
 
     @catch_exceptions_and_continue
-    def process_lockfiles(self, commit=None):
-
-        rev = commit or self.github_repo_branch[len("refs/heads/"):]
```
dustymabe: I think this parsing of the branch is important because when building the `raw.githubusercontent.com` URL below we use `{rev}` and using `refs/heads/testing-devel` here isn't going to work IIUC.

- https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/manifest-lock.x86_64.json

works

- https://raw.githubusercontent.com/coreos/fedora-coreos-config/refs/heads/testing-devel/manifest-lock.x86_64.json

doesn't

dustymabe:
one comment that should be easily fixed: https://github.com/coreos/fedora-coreos-releng-automation/pull/89#discussion_r395839926

coreos-koji-tagger/coreos_koji_tagger.py
jlebon: Good catch, thanks. Fixed!

jlebon:
Tested successfully in stg. Merging!

PR #90, opened 2020-03-27
coreos-meta-translator: add digitalocean

bgilbert:
Depends on https://github.com/coreos/fedora-coreos-stream-generator/pull/9.

### DIFF
--- a/coreos-meta-translator/trans.py
+++ b/coreos-meta-translator/trans.py
@@ -79,7 +79,7 @@ for f in files:
         # build the architectures dict
         arch_dict = {"media": {}}
         ensure_dup(input_, arch_dict, "ostree-commit", "commit")
-        generic_arches = ["aliyun", "aws", "azure", "exoscale", "qemu", "metal", "openstack", "vmware", "gcp"]
+        generic_arches = ["aliyun", "aws", "azure", "exoscale", "qemu", "metal", "openstack", "vmware", "gcp", "digitalocean"]
         for ga in generic_arches:
             if input_.get("images", {}).get(ga, None) is not None:
                 print(f"   - {ga}")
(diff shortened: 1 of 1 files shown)


coreos-meta-translator/trans.py
```diff
@@ -79,7 +79,7 @@ def artifact(i):
         # build the architectures dict
         arch_dict = {"media": {}}
         ensure_dup(input_, arch_dict, "ostree-commit", "commit")
-        generic_arches = ["aliyun", "aws", "azure", "exoscale", "qemu", "metal", "openstack", "vmware", "gcp"]
+        generic_arches = ["aliyun", "aws", "azure", "exoscale", "qemu", "metal", "openstack", "vmware", "gcp", "digitalocean"]
```
dustymabe: nit, looks like we were trying to go alphabetical originally, maybe we go back to that?

bgilbert:
Sorted the platform list, and also renamed a couple variables for clarity.

dustymabe:
ha nice. I was actually thinking to myself that "arch" was misleading, but didn't want to ask you to change that in this PR. Nice work!

lucab:
Uh, wait, this introduced a split brain definition of the `digitalocean` platform ID:
 * for Afterburn, it means a platform without DHCP requiring additional network bringup
 * for coreos-assembler, it means a platform with DHCP and no further special requirements

I fear this is going to be a blocker for providing first-class DO images (assuming they still need network bringup). I'm still reading through all the linked issues/PRs but I havent's seen yet those details of our plans. What is our strategy out of this?

bgilbert:
@lucab I don't think there's a conflict.  Afterburn runs successfully on `custom_ipxe` systems, and produces network information that matches DHCP.  So Afterburn's definition is fine, and once it supports driving NM, that can happen unconditionally.

lucab:
From out-of-band discussion: we are not totally sure, but this is likely going to be fine. If not, we'll come up with some additional way to let us do auto-detection here.

PR #91, opened 2020-04-09
coreos-koji-tagger: update ostree, include strace

dustymabe:


coreos-ostree-importer/Dockerfile
```diff
@@ -12,8 +12,8 @@ RUN dnf update -y
 RUN dnf -y install \
```
jlebon, line 12: Can drop the comment just before this line now.

dustymabe: done

PR #93, opened 2020-04-09
coreos-ostree-importer: remove permissions check for now

dustymabe:
We need to update a few pieces to set their umask to 0002 when modifying
files in the repos. robosignatory and pungi are two that still need to
be updated. For now let's just print the warnings and not error out.

### DIFF
--- a/coreos-ostree-importer/coreos_ostree_importer.py
+++ b/coreos-ostree-importer/coreos_ostree_importer.py
@@ -198,7 +198,8 @@ def assert_dirs_permissions(path: str):
             logger.warning(f"Directory {d} does not have gid=263!")
             founderror = True
     if founderror:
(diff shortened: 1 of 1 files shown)


dustymabe:
robosig PR: https://pagure.io/robosignatory/pull-request/45#

jlebon:
Is it safe to continue if the permissions aren't right though? Couldn't we end up with e.g. partially pulled commits for example?

dustymabe:
from looking at things from last night's run it looks like the only directories that were updated with the wrong permissions were under refs/heads. Since we're not writing to the same refs as the ones coming from pungi we should be safe for now.

dustymabe:
> robosig PR: https://pagure.io/robosignatory/pull-request/45#

and...

pungi PR: https://pagure.io/pungi/pull-request/1373

PR #99, opened 2020-04-17
coreos-ostree-importer: use newer ostree from koji

dustymabe:
Grab ostree-2020.3-3.fc31 from koji so we can hopefully get over
some concurrency issues. https://github.com/ostreedev/ostree/pull/2077

### DIFF
--- a/coreos-ostree-importer/Dockerfile
+++ b/coreos-ostree-importer/Dockerfile
@@ -8,11 +8,13 @@ ENV PYTHONUNBUFFERED=true
 RUN dnf update -y
 
 # Install needed rpms
+# Grab ostree from koji for now for https://github.com/ostreedev/ostree/pull/1984
 RUN dnf -y install \
         fedora-messaging \
         python-requests  \
-        ostree           \
-        strace
+        strace           \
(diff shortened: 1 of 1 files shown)


coreos-ostree-importer/Dockerfile
```diff
@@ -8,11 +8,13 @@ ENV PYTHONUNBUFFERED=true
...
 
 # Install needed rpms
+# Grab ostree from koji for now for https://github.com/ostreedev/ostree/pull/1984
 RUN dnf -y install \
         fedora-messaging \
         python-requests  \
-        ostree           \
-        strace
+        strace           \
+        https://kojipkgs.fedoraproject.org//work/tasks/9260/43499260/ostree-2020.3-3.fc31.x86_64.rpm \
```
jlebon, line 16: I think we want the links under `kojipkgs.fp.org/packages/`, no?

dustymabe: I have never used that particular URL.. I've always just used the URL directly from the koji task itself but I can swtich over to it (assuming it works).

dustymabe: updated!

PR #100, opened 2020-04-18
coreos-ostree-importer: fix send-import-request response validation

dustymabe:
This is a followup to c92a36e. We need to define the response variable.


dustymabe:
an example where this is currently failing: https://jenkins-fedora-coreos.apps.ci.centos.org/blue/organizations/jenkins/fedora-coreos%2Ffedora-coreos-fedora-coreos-pipeline/detail/fedora-coreos-fedora-coreos-pipeline/12332/pipeline

PR #101, opened 2020-04-21
importer: import send_request_and_wait_for_response function

jlebon:
Should fix the pipeline currently hitting:

> NameError: name 'send_request_and_wait_for_response' is not defined


dustymabe:
man between the two of us we'll eventually tackle this beast!

PR #108, opened 2020-06-09
coreos-meta-translator: Add Vultr to trans.py

dghubble:
* Consider Vultr generic images when translating meta to release


dghubble:
Borrowing from https://github.com/coreos/fedora-coreos-releng-automation/pull/82

PR #109, opened 2020-06-10
coreos-meta-translator: add gcp cloud launchable to release.json

zonggen:
As AMIs have their own field `images` under the `aws` key, I think it might make sense for GCP to follow the same rule, adding an `image` field under the `gcp` key and rename the original `image` key to `name`.

If this is reasonable, it should be followed by updating `fedora-coreos-stream-generator` to support this PR. Finally, download page could read the GCP cloud launchable metadata from e.g. `stable.json` instead of `meta.json`

Related: https://github.com/coreos/fedora-coreos-tracker/issues/494

### DIFF
--- a/coreos-meta-translator/trans.py
+++ b/coreos-meta-translator/trans.py
@@ -102,6 +102,13 @@ for f in files:
                     "image": ami_dict["hvm"]
                 }
 
+        # GCP specific additions
+        if input_.get("gcp", None) is not None:
+            arch_dict["media"]["gcp"] = arch_dict["media"].get("gcp", {})
+            arch_dict["media"]["gcp"]["image"] = arch_dict["media"]["gcp"].get("image", {})
+            arch_dict["media"]["gcp"]["image"].update(input_.get("gcp", {}))
+            arch_dict["media"]["gcp"]["image"]["name"] = arch_dict["media"]["gcp"]["image"].pop("image")
+
         # metal specific additions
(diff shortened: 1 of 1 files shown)


bgilbert:
https://github.com/coreos/fedora-coreos-tracker/issues/494#issuecomment-642324355

zonggen:
Closing the issue as we inline to not exposing the GC URL, thanks for the link :+1:

dustymabe:
hey @zonggen - I think the only thing we agreed on so far is to not expose the URL - so I think we'd still want the project/family/name.

dustymabe:
So, IIUC, something like:

```
"gcp":{
   "artifacts":{
      "tar.gz":{
         "disk":{
            "location":"https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/31.20200517.3.0/x86_64/fedora-coreos-31.20200517.3.0-gcp.x86_64.tar.gz",
            "signature":"https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/31.20200517.3.0/x86_64/fedora-coreos-31.20200517.3.0-gcp.x86_64.tar.gz.sig",
            "sha256":"e7a766cbba51fc35cb55d471c42ecc824519c6a3bd8a95577749676aeebf9b9c"
         }
      }
   },
   "image":{
      "project":"fedora-coreos-cloud",
      "family":"fedora-coreos-stable",
      "name":"fedora-coreos-31-20200517-3-0-gcp-x86-64"
   }
}
```

PR #112, opened 2020-06-15
coreos-meta-translator: add gcp cloud launchable to release.json

zonggen:
As AMIs have their own field `images` under the `aws` key, I think it might make sense for GCP to follow the same rule, adding an `image` field under the `gcp` key and rename the original `image` key to `name`.

If this is reasonable, it should be followed by updating `fedora-coreos-stream-generator` to support this PR. Finally, download page could read the GCP cloud launchable metadata from e.g. `stable.json` instead of `meta.json`

Related: https://github.com/coreos/fedora-coreos-tracker/issues/494

### DIFF
--- a/coreos-meta-translator/trans.py
+++ b/coreos-meta-translator/trans.py
@@ -102,6 +102,15 @@ for f in files:
                     "image": ami_dict["hvm"]
                 }
 
+        # GCP specific additions
+        if input_.get("gcp", None) is not None:
+            arch_dict["media"]["gcp"] = arch_dict["media"].get("gcp", {})
+            arch_dict["media"]["gcp"]["image"] = arch_dict["media"]["gcp"].get("image", {})
+            arch_dict["media"]["gcp"]["image"].update(input_.get("gcp", {}))
+            arch_dict["media"]["gcp"]["image"]["name"] = arch_dict["media"]["gcp"]["image"].pop("image")
+            # remove the url as we haven't decided to expose that information publicly yet
+            arch_dict["media"]["gcp"]["image"].pop("url")
+
         # metal specific additions
         arch_dict["media"]["metal"] = arch_dict["media"].get("metal", {})
         arch_dict["media"]["metal"]["artifacts"] = arch_dict["media"]["metal"].get("artifacts", {})


zonggen:
Continuation of https://github.com/coreos/fedora-coreos-releng-automation/pull/109, somehow GH does not allow me to reopen the PR, so moved to here.

coreos-meta-translator/trans.py
```diff
@@ -102,6 +102,15 @@ def artifact(i):
...
                 }
 
+        # GCP specific additions
+        if input_.get("gcp", None) is not None:
+            arch_dict["media"]["gcp"] = arch_dict["media"].get("gcp", {})
+            arch_dict["media"]["gcp"]["image"] = arch_dict["media"]["gcp"].get("image", {})
+            arch_dict["media"]["gcp"]["image"].update(input_.get("gcp", {}))
+            if arch_dict["media"]["gcp"]["image"].get("image", None) is not None:
+                arch_dict["media"]["gcp"]["image"]["name"] = arch_dict["media"]["gcp"]["image"].pop("image", None)
+            arch_dict["media"]["gcp"]["image"].pop("url", None)
```
dustymabe, line 112:
```suggestion
            # remove the url as we haven't decided to expose that information publicly yet
            arch_dict["media"]["gcp"]["image"].pop("url", None)
```

zonggen: Updated

coreos-meta-translator/trans.py
```diff
@@ -102,6 +102,16 @@ def artifact(i):
                     "image": ami_dict["hvm"]
                 }
 
+        # GCP specific additions
+        if input_.get("gcp", None) is not None:
+            arch_dict["media"]["gcp"] = arch_dict["media"].get("gcp", {})
+            arch_dict["media"]["gcp"]["image"] = arch_dict["media"]["gcp"].get("image", {})
+            arch_dict["media"]["gcp"]["image"].update(input_.get("gcp", {}))
+            if arch_dict["media"]["gcp"]["image"].get("image", None) is not None:
```
jlebon, line 110: Should this be conditional? Feels like it should be an error if `image` wasn't in the input, no? IOW, can we drop this and instead just let `pop` error out if it's missing?

zonggen: Emm.. I think this condition will always be true, since `arch_dict["media"]["gcp"]["image"].update(input_.get("gcp", {}))` made sure it is at least a `{}`. Anyway this is redundant, dropping this

zonggen: Updated

coreos-meta-translator/trans.py
```diff
@@ -102,6 +102,15 @@ def artifact(i):
                     "image": ami_dict["hvm"]
                 }
 
+        # GCP specific additions
+        if input_.get("gcp", None) is not None:
+            arch_dict["media"]["gcp"] = arch_dict["media"].get("gcp", {})
+            arch_dict["media"]["gcp"]["image"] = arch_dict["media"]["gcp"].get("image", {})
+            arch_dict["media"]["gcp"]["image"].update(input_.get("gcp", {}))
+            arch_dict["media"]["gcp"]["image"]["name"] = arch_dict["media"]["gcp"]["image"].pop("image", None)
```
jlebon, line 110: My suggestion is to make it a hard error so we don't silently continue if `image` is missing from `meta.json`:

```suggestion
            arch_dict["media"]["gcp"]["image"]["name"] = arch_dict["media"]["gcp"]["image"].pop("image")
```

zonggen: Oh I see, and the previous `pop("image")` too. Updating

PR #113, opened 2020-06-18
coreos-ostree-importer: use new OSTree with EOPNOTSUPP fix

jlebon:
https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/18

Drop when it hits stable:
https://bodhi.fedoraproject.org/updates/FEDORA-2020-4ba7f57a90


dustymabe:
LGTM

PR #118, opened 2020-07-27
coreos-koji-tagger: detonate pod if we get a koji auth error

dustymabe:
We've tried to work around the issue in #70 several ways including
d46a3a1, a90d53d, and 286612b. Nothing seems to be working and we are
currently still at https://github.com/coreos/fedora-coreos-releng-automation/issues/70#issuecomment-600820800.
Restarting the pod seems to work well enough so let's just self destruct
if we detect this failure case. The new pod that gets created will
do a scan of the repos and pick up right where the old one left off
so this should work just fine.

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -15,7 +15,6 @@ import yaml
 import time
 
 from koji_cli.lib import watch_tasks
-from tenacity import retry, wait_fixed, stop_after_attempt, retry_if_exception_type
 
 # Set local logging 
 logger = logging.getLogger(__name__)
@@ -194,6 +193,10 @@ def catch_exceptions_and_continue(func):
     def wrapper(*args, **kwargs):
         try:
             return func(*args, **kwargs)
+        except koji.AuthError as e:
+            logger.error('Received koji.AuthError from koji. Exiting so a new pod will get created.')
+            logger.error('See https://github.com/coreos/fedora-coreos-releng-automation/issues/70')
+            raise
         except Exception as e:
             logger.error('Caught Exception!')
             logger.error('###################################')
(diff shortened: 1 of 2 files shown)


coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -447,22 +447,22 @@ def process_lockfiles(self, rev):
                 watch_tasks(self.koji_client, [task], poll_interval=10)
                 logger.info('Dist-repo task has finished')
 
-    # retry to login every 30s for 15m before giving up
-    # https://github.com/coreos/fedora-coreos-releng-automation/issues/70
-    @retry(retry=retry_if_exception_type(koji.AuthError), wait=wait_fixed(30), stop=stop_after_attempt(30))
```
bgilbert, line 452: We can drop the tenacity dependency now.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -447,22 +447,22 @@ def process_lockfiles(self, rev):
...
                 return
         except koji.AuthError as e:
-            logger.info('Received koji.AuthError from koji. Re-attempting login.')
-            pass
-        # Login!
-        principal = find_principal_from_keytab(self.keytab_file)
-        self.koji_client.gssapi_login(principal, self.keytab_file)
+            logger.error('Received koji.AuthError from koji. Exiting so a new pod will get created.')
+            logger.error('See https://github.com/coreos/fedora-coreos-releng-automation/issues/70')
+            sys.exit(1)
```
bgilbert, line 465: Rather than doing this here, why not let `koji.AuthError` propagate, and special-case it in `catch_exceptions_and_continue`?

dustymabe: just pushed up a new commit, is that what you mean? If so I'll squash the commit into the previous commit.

dustymabe:
ok squashed that last commit, now with just one commit with an appropriate commit message

PR #119, opened 2020-08-07
coreos-release-notes-generator: init script for generating release notes

zonggen:
This script generates Fedora CoreOS release note from pre-mature release note snippets. Takes two arguments: `--config-dir` and `--output-dir`.

The `config_dir` is the directory where a `'release-notes.d` directory resides, typically a `fedora-coreos-config` directory, which has pre-mature release notes yaml snippets.

The `output_dir` is the directory where a `release-notes.yaml` resides. If there's no such file then a new file will be created. If a `release-notes.yaml` already exists, then checks if the `next-release` field is present in the file. It inserts the `next-release` field to the front of the file if none is present or updates the existing `next-release` field if one already exists.

Related: https://github.com/coreos/fedora-coreos-tracker/issues/194

### DIFF
--- a/coreos-release-notes-generator/README.md
+++ b/coreos-release-notes-generator/README.md
@@ -1,0 +1,82 @@
+# coreos-release-notes-generator
+
+`coreos-release-notes-generator` generates `release-notes.yaml` from `release-notes.d` yaml snippets using the specified build id using `--build-id` option.
+
+## Options
+ - `--build-id`: build id of the latest release
+ - `--config-dir`: path to the directory where `release-notes.d/` lives
+ - `--release-notes-file`: path to the input `release-notes.yaml` file for update
+ - `--output-dir`: output directory for `release-notes.yaml`
+ - `--json`: output JSON format instead of YAML for easier consumption by Fedora CoreOS website (https://getfedora.org/en/coreos?stream=stable)
+
+As an example, assuming following structure:
+```
+.
+├── fedora-coreos-config
+└── fedora-coreos-releng-automation
+
+```
+
+## Generate a new `release-notes.yaml`
+
+To generate a new `release-notes.yaml`, simply omit `--release-notes.yaml`:
+```
+$ ./release-notes-generator.py --config-dir ../../fedora-coreos-config/ \
+    --build-id 32.20200817.2.1
+
+- 32.20200817.2.1:
+    coreos-installer:
+    - subject: installer 4
+    - body: installer body 1
+      subject: installer 1
+    - subject: installer 2
+    - subject: installer 3
+    ignition:
+    - body: ignition body 1
+      subject: ignition 1
+    - subject: igntiion 2
+    miscellaneous:
+    - subject: misc 2
+    - subject: misc 1
+
+```
+
+## Update existing `release-notes.yaml`
+
+To update existing `release-notes.yaml`:
+```
+$ cat ../../fedora-coreos-config/release-notes.yaml
+- 32.20200801.2.1:
+    afterburn:
+    - body: afterburn body 1
+      subject: afterburn 1
+    - subject: afterburn 2
+    miscellaneous:
+    - subject: misc 1
+
+$ ./release-notes-generator.py --release-notes-file ../../fedora-coreos-config/release-notes.yaml \
+    --config-dir ../../fedora-coreos-config/ \
+    --build-id 32.20200817.2.1
+- 32.20200817.2.1:
+    coreos-installer:
+    - subject: installer 4
+    - body: installer body 1
+      subject: installer 1
+    - subject: installer 2
+    - subject: installer 3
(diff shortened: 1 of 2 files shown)


dustymabe:
Thanks @zonggen for working on this! A few questions/comments:

- Can we add a nice README to this directory that explains how to use it and also what format files it expects? Some of the information from https://github.com/coreos/fedora-coreos-config/pull/550 was useful in putting the pieces together.
- Can we break up the file so that it uses functions more. I would break it up so that each logical piece is in its own function and there is a `main` function that gets run (for inspiration see https://github.com/coreos/fedora-coreos-releng-automation/blob/master/coreos-ostree-importer/coreos_ostree_importer.py#L369). Don't worry about breaking it down too much, just getting it set up that way should help it flow a bit better.

zonggen:
@dustymabe Thanks for reviewing! Will cleanup the code and add a README for documentation. Also will update with the updated schema in https://github.com/coreos/fedora-coreos-config/pull/550

zonggen:
Removed `next-release` intermediate logic and added documentations.

Currently working on adding a list of file names such that no duplication between multiple `testing` releases, since we need yaml snippets to accumulate under `release-notes.d/` for `stable`.

coreos-release-note-generator/README.md
```diff
@@ -0,0 +1,82 @@
+# coreos-release-note-generator
```
dustymabe, line 1: Should we rename to `coreos-release-notes-generator`?

coreos-release-note-generator/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+`coreos-release-note-generator` generates `release-notes.yaml` from `release-notes.d` yaml snippets using the specified build id using `--build-id` option.
+
+## Options
+ - `--build-id`: build id of the latest release note
```
dustymabe, line 6:
```suggestion
 - `--build-id`: build id of the latest release
```

coreos-release-note-generator/release-note-generator.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if not os.path.exists(os.path.join(args.config_dir, 'release-notes.d')):
+        raise Exception(
+            "release-notes.d does not exist under {}".format(args.config_dir))
+
+    snippet_yaml_list = glob.glob(os.path.join(
+        args.config_dir, 'release-notes.d/*.yaml'))
+    if len(snippet_yaml_list) == 0:
+        print("release-notes.d/ does not contain any yaml snippets under '{}'".format(args.config_dir))
+        exit(0)
```
dustymabe, line 76: Can we return here instead of exiting in the middle of a function?

coreos-release-note-generator/release-note-generator.py
```diff
@@ -0,0 +1,167 @@
...
+
+    snippet_dict = dict()
+    for snippet_yaml in snippet_yaml_list:
+        with open(snippet_yaml, 'r') as f:
+            snippet = yaml.load(f, Loader=yaml.FullLoader)
+            for item in snippet:
+                note = {'subject': item.get(
+                    'subject', ''), 'body': item.get('body', '')}
+                # purposely avoid item.get('component', '') to error out if the component key does not exist
+                project_name = item['component']
```
dustymabe, line 86: instead of using `project` throughout this file should we just call it `component` instead?

coreos-release-note-generator/release-note-generator.py
dustymabe: Also, we should probably consider the case where we haven't added any release-notes.d/ files for a particular release. I guess in that case we should just put an empty entry in the list saying there are not release notes or something.

coreos-release-note-generator/README.md
zonggen: Agreed :+1:

coreos-release-note-generator/release-note-generator.py
zonggen: Or, we could just merge this *list* of *build_id: release_notes* items into a single dictionary so that we could query whether a release note exists for a release just by checking if the key *build_id* exists. This could be done when we are converting to JSON format for website consumption, on the other hand I think it makes sense to use YAML file for human readability on other places like `fedora-coreos-config`.

coreos-release-note-generator/release-note-generator.py
zonggen: Agreed :+1:

dustymabe:
one small nit and then we can merge:

Change commit subject line to `coreos-release-notes-generator: init script for generating release notes`

zonggen:
Fixed commit subject, thanks for pointing that out!

dustymabe:
LGTM - I think we can merge this and do fixups for things we find in a followup PR.

PR #122, opened 2020-08-24
config-bot: handle flakes from GitHub when fetching

jlebon:
First, don't implicitly fetch all the latest commits when acquiring a
`Git` context; let's make it a separate `fetch()` function.

Second, handle errors from fetching that could be due to networking
flakes or e.g. GitHub being down. We want `config-bot` to just absorb
all transient errors and retry.

### DIFF
--- a/config-bot/main
+++ b/config-bot/main
@@ -130,6 +130,13 @@ async def do_sync(base_url, method, stream, build):
             return False
 
     async with git:
+        try:
+            git.fetch()
+        except Exception as e:
+            # flaked while fetching from GitHub? just ignore, we'll retry
+            logging.error(f"Got exception during fetch: {e}")
+            return False
+
         git.checkout(stream)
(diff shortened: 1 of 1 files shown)


config-bot/main
```diff
@@ -329,7 +350,6 @@ class Git:
     async def __aenter__(self):
         logging.info("acquiring lock")
         await self._lock.acquire()
-        self.cmd('fetch', 'origin', '--prune', '+refs/heads/*:refs/heads/*')
```
dustymabe, line 332: I don't have a lot of knowledge of the async code so I'll ask:

This used to happen after "acquiring lock" - does this PR have any implications there?

jlebon: It's still done under lock; it's kept while using the `Git` context.

jlebon: That's a good sanity-check though!

PR #123, opened 2020-08-26
config-bot: a few tweaks following migration to OCP4 cluster

jlebon:
```
commit 60e4489d8c53c4abbcb25595a0697220ad76be6d
Date:   Wed Aug 26 15:40:42 2020 -0400

    config-bot/manifest: use same GitHub coreosbot secret as pipeline

    Since we're deployed in the same namespace anyway, might as well reuse
    the same secret.

commit bdfde294b4758bf88ef093f13982414dd6159e58
Date:   Wed Aug 26 15:41:11 2020 -0400

    config-bot/README: add details about deploying to OpenShift

    Pretty straight-forward, but just to have it written down somewhere.
```

### DIFF
--- a/config-bot/manifest.yaml
+++ b/config-bot/manifest.yaml
@@ -66,13 +66,13 @@ objects:
             - name: config-bot
               image: config-bot
               volumeMounts:
-                - name: github-token-mount
+                - name: github-coreosbot-token
                   mountPath: /var/run/secrets/coreos.fedoraproject.org/github-token
                   readOnly: true
           volumes:
-          - name: github-token-mount
+          - name: github-coreosbot-token
             secret:
-              secretName: config-bot-github-token
+              secretName: github-coreosbot-token
       triggers:
         - type: ConfigChange
         - type: ImageChange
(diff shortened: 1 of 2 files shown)


config-bot/manifest.yaml
```diff
@@ -66,13 +66,13 @@ objects:
...
               volumeMounts:
-                - name: github-token-mount
+                - name: github-coreosbot-token
                   mountPath: /var/run/secrets/coreos.fedoraproject.org/github-token
                   readOnly: true
           volumes:
-          - name: github-token-mount
+          - name: github-coreosbot-token
             secret:
-              secretName: config-bot-github-token
```
lucab, line 75: Did we already (or plan to) invalidate this old one?

lucab (approved):

config-bot/manifest.yaml
jlebon: It was actually the same token string, just under different secrets. They're both there in the old cluster; in the new cluster, they already share the same secret.

jlebon:
Thanks for the review!

PR #125, opened 2020-10-30
scripts: add broadcast-fedmsg.py

jlebon:
This is a thin wrapper around the new `broadcast_fedmsg` function:
https://github.com/coreos/coreos-assembler/pull/1821

This is the script to which the Fedora CoreOS pipelines will call out.

### DIFF
--- a/scripts/broadcast-fedmsg.py
+++ b/scripts/broadcast-fedmsg.py
@@ -1,0 +1,109 @@
+#!/usr/bin/python3
+
+'''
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1821

dustymabe (approved):
LGTM. Optional would be to add a description in a comment at the top of the file that explains its usefulness a bit.

PR #126, opened 2020-11-10
broadcast-fedmsg: put build dir URL in build.state.change msg

jlebon:
Otherwise it will require services which want to make use of these
fedmsgs to hardcode knowledge of our bucket layout. Don't make it
required so we can ratchet it in.


dustymabe (approved):
LGTM

PR #127, opened 2020-12-18
config-bot: propagate files from testing-devel to rawhide

jlebon:
We'll try for now to keep using testing-devel as the source config ref
for rawhide builds. We can adapt in the future if it gets too tricky.


dustymabe (approved):
LGTM

PR #130, opened 2021-02-18
config-bot: propagate files from testing-devel to branched

jlebon:
Let's just stay with the theme of `testing-devel` being the source ref
for now.


cgwalters (approved):
Superficial LGTM

PR #134, opened 2021-03-26
config-bot: don't sync image.yaml

jlebon:
We now have `image-base.yaml` and `image.yaml`, where the latter is only
meant for a specific stream, while the former is shared. For more
information, see:

https://github.com/coreos/fedora-coreos-tracker/issues/292#issuecomment-806099932
https://github.com/coreos/coreos-assembler/pull/2096
https://github.com/coreos/fedora-coreos-config/pull/908

### DIFF
--- a/config-bot/config.toml
+++ b/config-bot/config.toml
@@ -38,6 +38,7 @@ target-refs = [
 skip-files = [
     'manifest.yaml',
     'manifest-lock.*',
+    'image.yaml',
 ]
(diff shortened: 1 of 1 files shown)


jlebon:
Hold until https://github.com/coreos/fedora-coreos-config/pull/908 is merged and the files are synced to the `next-devel` branch.

sohankunkerkar (approved):

PR #137, opened 2021-06-02
config-bot/manifest.yaml: use separate token for config-bot

jlebon:
All config-bot needs is cloning and pushing. Let's use a separate token
for that from the more powerful github-coreosbot-token.


dustymabe (approved):
LGTM

PR #140, opened 2021-07-14
coreos-koji-tagger: watch/tag override lockfile from the branched branch

dustymabe:
We'll be using `branched` soon for f35 (after branching occurs). Let's
watch those lockfiles too.

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -44,7 +44,7 @@ COREOS_KOJI_USER = 'coreosbot'
 
 # XXX: should be in config file
 DEFAULT_GITHUB_REPO_FULLNAME = 'coreos/fedora-coreos-config'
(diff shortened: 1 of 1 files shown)


coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -44,7 +44,7 @@
 
 # XXX: should be in config file
 DEFAULT_GITHUB_REPO_FULLNAME = 'coreos/fedora-coreos-config'
-DEFAULT_GITHUB_REPO_BRANCHES = 'refs/heads/testing-devel refs/heads/next-devel refs/heads/rawhide'
+DEFAULT_GITHUB_REPO_BRANCHES = 'refs/heads/testing-devel refs/heads/next-devel refs/heads/rawhide  refs/heads/branched'
```
jlebon, line 47: Hmm, maybe at this point we should just trim the `refs/heads/` and account for that in the (AFAICT) one spot that cares in `process_message`. But meh... obviously fine as is too!

jlebon (approved):

PR #142, opened 2021-07-30
coreos-ostree-importer: support passing in target architecture

dustymabe:
This will allow us to send messages to import aarch64 commits
even if we're running in a x86_64 environment.

### DIFF
--- a/coreos-ostree-importer/send-ostree-import-request.py
+++ b/coreos-ostree-importer/send-ostree-import-request.py
@@ -28,7 +28,7 @@ OSTREE_IMPORTER_REQUEST_TIMEOUT_SEC = 15 * 60
 
 def parse_args():
     parser = argparse.ArgumentParser()
-    parser.add_argument("--build", help="Build ID", default="latest")
+    parser.add_argument("--build", help="Build ID", required=True)
     parser.add_argument(
         "--arch", help="target architecture", default=get_basearch()
     )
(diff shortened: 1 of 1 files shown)


coreos-ostree-importer/send-ostree-import-request.py
```diff
@@ -43,6 +44,10 @@ def parse_args():
         required=True,
         help="bucket and prefix to S3 builds/ dir",
     )
+    parser.add_argument(
+        "--arch",
+        help="target architecture",
+    )
```
jlebon, line 50: Minor/optional: seems like this would fit fine on one line. Other ones are split to satisfy pep8.

jlebon (approved):

dustymabe:
slightly re-worked the PR

coreos-ostree-importer/send-ostree-import-request.py
```diff
@@ -53,26 +57,24 @@ def parse_args():
 
 
 def send_ostree_import_request(args):
-    buildid = args.build
-    build = GenericBuildMeta(build=buildid)
+    build = GenericBuildMeta(build=args.build, basearch=args.arch)
```
jlebon, line 60: This isn't new, though I think here we should have e.g.

```python
buildid = build['buildid']
```

and use that below. Because it doesn't seem appropriate in a signing context to just pass through `latest` unresolved if `--build` wasn't actually passed. We should probably also just remove that default from the argparse argument.

Anyway, since it's not new, OK to defer.

jlebon (approved):

coreos-ostree-importer/send-ostree-import-request.py
dustymabe: ok tried to address this in a new commit - pushed

jlebon (approved):

PR #144, opened 2021-09-10
coreos-koji-tagger: bump signing timeout to 30 mins

jlebon:
We just hit the case where a lot of packages needed to be tagged, and
there was genuine progress being made but the timer ran out. So then I
had to manually run `koji dist-repo`.

Let's just bump the timeout to 30 minutes to catch this.

A smarter method we could follow-up with on this would be to check if
the difference is getting smaller and reset the timeout if so.


dustymabe:
My experience is that when it's blocked because of large things going through the queue, it's blocked for a few hours. IOW I don't think this will help most of the time we see the problem, but it's worth a try.. Would love to somehow get priority in robosignatory for our stuff.

dustymabe (approved):
LGTM

Issue #145, opened 2021-10-07
adapt to new ostree-container ociarchive


cgwalters:
For https://github.com/coreos/fedora-coreos-tracker/issues/812

https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/job/fedora-coreos/job/fedora-coreos-fedora-coreos-pipeline/2919/
```
Exception: Importing failed: Command '['ostree', '--repo=/tmp/tmp6cohmf6g', 'show', '469e3824769141206cd483e351661a53855c60102c4ef04a34b1de7cb20b0089', '--print-metadata-key=version']' returned non-zero exit status 1.
```
This is *another* copy of code like https://github.com/coreos/coreos-assembler/pull/2487

dustymabe:
hmm is `--mode=bare-user` right? the tarballs we have today are `mode=archive-z2`

PR #146, opened 2021-10-07
coreos-ostree-importer: support OCI archives

jlebon:
FCOS rawhide is now producing archives in the new OCI format. We have to
use `rpm-ostree ex-container import` to extract those.

Closes: #145


dustymabe:
WDYT of reworking slightly the code above this to be a little less misleading in the case it's not a tarball?

```
filename = "ostree.tar"
filepath = os.path.join(tmpdir, filename)
```

It might suffice to just change it from `ostree.tar` to `ostreearchive` (with no suffix).

PR #148, opened 2022-01-27
config-bot: add README instructions to create secret token and drop BuildConfig GitHub trigger

jlebon:
See individual commits.

### DIFF
--- a/config-bot/README.md
+++ b/config-bot/README.md
@@ -64,10 +64,3 @@ To deploy:
 ```
 oc new-app --file=manifest.yaml
 ```
-
-Copy the generated GitHub webhook secret, and substitute it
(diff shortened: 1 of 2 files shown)


dustymabe:
> It relies on webhooks sent to the API server, which is not externally
accessible in the new cluster.

Should we open a ticket to see if they can figure out a way to accommodate us?

dustymabe (approved):

jlebon:
> > It relies on webhooks sent to the API server, which is not externally
> > accessible in the new cluster.
> 
> Should we open a ticket to see if they can figure out a way to accommodate us?

Will do!

jlebon:
https://pagure.io/fedora-infrastructure/issue/10521

dustymabe:
Thank you!

PR #149, opened 2022-02-10
scripts/promote-config: add `LOCAL` for local hacking

jlebon:
If e.g. you're hacking on `testing-devel` locally and then want to test
`next-devel` with those changes as if they were propagated by the bot,
you can use this.

### DIFF
--- a/scripts/promote-config.sh
+++ b/scripts/promote-config.sh
@@ -21,8 +21,12 @@ main() {
     local src_branch=$1; shift
     local fetch_head head
 
-    git fetch https://github.com/coreos/fedora-coreos-config "${src_branch}"
-    fetch_head=$(git rev-parse FETCH_HEAD)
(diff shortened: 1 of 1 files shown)


dustymabe:
hmm. What's the use case here? basically where you *would* cherry pick individual commits now you can use this tool to do it all for you?

dustymabe (approved):

jlebon:
> hmm. What's the use case here? basically where you _would_ cherry pick individual commits now you can use this tool to do it all for you?

I'm using this to test the iptables-nft migration. I want to test https://github.com/coreos/fedora-coreos-config/pull/1460 as if https://github.com/coreos/fedora-coreos-config/pull/1324 had merged and propagated without actually having to merge https://github.com/coreos/fedora-coreos-config/pull/1324 first.

PR #154, opened 2022-06-10
Revert "config-bot: temporarily skip syncing platforms.yaml"

bgilbert:
Once `platforms.yaml` promotes to `testing-devel`, we will no longer need it to diverge across our development branches.

Reverts #153.


dustymabe (approved):

jlebon:
Cool with this, but in the future I think I'd prefer tackling this sort of thing differently. E.g. `image.yaml` knob instead, or designing from the start a `platform-base.yaml` and `platform.yaml` split like `image.yaml` does.

PR #155, opened 2022-08-04
config-bot/manifest.yaml: remove unnecessary fields

jlebon:
Those fields are only useful if instantiating the template itself, which
we're not doing here.

This is similar to what was done in:
https://github.com/coreos/fedora-coreos-pipeline/pull/582

My main motivation though is to get a commit in so that we can test that
https://pagure.io/fedora-infrastructure/issue/10603 is resolved.

### DIFF
--- a/config-bot/manifest.yaml
+++ b/config-bot/manifest.yaml
@@ -1,12 +1,7 @@
 apiVersion: v1
 kind: Template
-labels:
-  app: config-bot
 metadata:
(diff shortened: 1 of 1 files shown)


jlebon:
If this test works, I'll modify all the other buildconfig webhooks on this repo to use SSL verification.

dustymabe (approved):
LGTM

jlebon:
Cool, it works! Other webhooks updated.

PR #158, opened 2022-10-11
coreos-ostree-importer: drop support for tarballs

jlebon:
It's been a while now since we've switch to OCI archives for all our streams. Let's drop support for the previous tarball path.


dustymabe (approved):
LGTM

PR #166, opened 2023-01-13
fedora-ostree-pruner: enable pruning the prod OSTree repo

dustymabe:
Now that the 2022 end of year holidays are over let's go ahead and prune!

### DIFF
--- a/fedora-ostree-pruner/fedora-ostree-pruner
+++ b/fedora-ostree-pruner/fedora-ostree-pruner
@@ -250,8 +250,7 @@ def main():
 
     # Prune the compose and the prod repo based on policy
     prune_compose_repo(args.test)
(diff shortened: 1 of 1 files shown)


dustymabe:
Added a second commit here:

```
commit 2771497cc193b3693a88169f30cb6d54aa13dce1
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Fri Jan 13 15:53:17 2023 -0500

    fedora-ostree-pruner: don't run on start in --loop mode
    
    We don't want the pruner to immediately run every time the container
    is started in OpenShift (which could happen for various reasons).
    
    In that case (the --loop case) we still only want it to run on the
    weekend (we chose Saturday).

```

jlebon (approved):

PR #168, opened 2023-01-31
use dumb-init for the coreos-ostree-importer and fedora-ostree-pruner

dustymabe:
Sometimes we use the pods for the importer and pruner to inspect and fix issues with the OSTree repos since it's the most efficient way for us to gain read/write access to the repos. Recently I was restoring some data that was lost in a recent prune [1] and the `ostree pull` operation I was running failed with an obscure error message:

```
error: Commit 1eb251bced7652f2f486c15447a7bf00238ef0ea172b4f214d2684ecfbeb2c40:
GPG: GPG: Failed to import key: GPGME: System error w/o errno
```

It turns out the many gpg processes that get forked during a pull operation to verify the signatures for the commits don't get cleaned up:

```
1000720+   11737       1  0 14:59 pts/4    00:00:00 [gpg] <defunct>
1000720+   11740       1  0 14:59 pts/4    00:00:00 [gpg] <defunct>
1000720+   11743       1  0 14:59 pts/4    00:00:00 [gpg] <defunct>
1000720+   11746       1  0 14:59 pts/4    00:00:00 [gpg] <defunct>
...
```

Let's use `dumb-init` to reap these defunct processes.

[1] https://github.com/fedora-silverblue/issue-tracker/issues/405#issuecomment-1407794147

### DIFF
--- a/fedora-ostree-pruner/Dockerfile
+++ b/fedora-ostree-pruner/Dockerfile
@@ -7,8 +7,8 @@ ENV PYTHONUNBUFFERED=true
 # Get any latest updates since last container spin
 RUN dnf update -y
 
-# Install ostree
-RUN dnf -y install ostree
+# Install ostree/dumb-init
+RUN dnf -y install ostree dumb-init
 
 # Configure a umask of 0002 which will allow for the group permissions
 # to include write for newly created files. We need this because we'd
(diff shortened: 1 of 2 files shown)


fedora-ostree-pruner/Dockerfile
```diff
@@ -7,8 +7,8 @@ ENV PYTHONUNBUFFERED=true
 # Get any latest updates since last container spin
 RUN dnf update -y
 
-# Install ostree
-RUN dnf -y install ostree
+# Install ostree/dumb-init
```
jlebon, line 10: Aside: this comment should probably be removed.

jlebon (approved):

fedora-ostree-pruner/Dockerfile
dustymabe: because it's superfluous?

jlebon: Yeah, exactly.

PR #171, opened 2023-05-03
config-bot: add experimental branch and support branch-level skip files

jlebon:
With the new experimental stream, we want to have more freedom in what
files go in there without having to also pollute `testing-devel`.

Add support for a `.coreos.skip-files` file which will tell config-bot
to not clobber a specific set of files on that ref. This augments the
global-level `skip-files` list.

Related: https://github.com/coreos/fedora-coreos-tracker/issues/1446


cgwalters (approved):
Seems sane to me

jlebon:
OK, let's hold this until we have the experimental branch up so it doesn't try to sync to a missing branch. Want to add `.coreos.skip-files` to your commit and create the branch? (And also the `prod` flip.)

dustymabe (approved):
LGTM

PR #178, opened 2023-08-22
scripts/promote-config: strip warn: out of denylist on promotion

dustymabe:
warn: true was added in https://github.com/coreos/coreos-assembler/pull/3539 and can be used to make test failures non-fatal in the pipeline. Let's strip those from the denylist too on promotion.

### DIFF
--- a/scripts/promote-config.sh
+++ b/scripts/promote-config.sh
@@ -36,9 +36,9 @@ main() {
     # except for manifest.yaml
     git checkout -- manifest.yaml
 
-    # also strip out the snoozes in the denylist because we don't want
-    # changes in the executed tests over time for production streams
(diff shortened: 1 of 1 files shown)


nikita-dubrovskii:
LGTM

scripts/promote-config.sh
```diff
@@ -36,9 +36,9 @@ main() {
     # except for manifest.yaml
     git checkout -- manifest.yaml
 
-    # also strip out the snoozes in the denylist because we don't want
-    # changes in the executed tests over time for production streams
-    sed -E -i 's/^(\s+)(snooze:\s+.*)/\1# \2 (disabled on promotion)/' kola-denylist.yaml
```
travier, line 41: Is there any specific reason `kola-denylist.yaml` was removed here?

It looks likes this is the reason for today's promote config failures: https://github.com/coreos/fedora-coreos-streams/actions/runs/6071168339/job/16468658157

Issue #181, opened 2023-12-01
coreos-koji-tagger: tag packages from rawhide builds


jlebon:
As part of Bodhi testing, we want to make sure we test against a known working baseline. We have this in the locked development streams, but not rawhide. Rather than adding lockfiles to rawhide, an intermediate approach is to:
1. have coreos-koji-tagger tag the packages from successful rawhide builds (from the generated lockfiles in the build dir in S3)
2. have the `test-override` job use the exact same lockfile as the last successful rawhide build and at the exact same fedora-coreos-config repo commit that was built

jlebon:
@dustymabe When we discussed this, I mentioned that ideally this doesn't go into the pool, but I think actually that's not going to be an issue because we *will* have a lockfile (the auto-generated one from the last build), so rpm-ostree will allow getting locked pakages from the pool.

PR #183, opened 2024-01-17
coreos-koji-tagger: tag packages from rawhide builds into pool

jlebon:
For the Bodhi testing work, when testing a rawhide update, we want to
test against a known working rawhide build. To do this, we want to base
our build on the last working rawhide combination. We do this by reusing
the generated lockfile from that build and the git commit of the config
repo.

But for this to work, we need to make sure that we have access to all
the RPMs in that build.

This patch teaches coreos-koji-tagger to listen to the
`coreos.stream.release` fedmsg that the release job emits to know
when a new rawhide build is available; it then downloads the generated
lockfiles and tags in the packages.

Closes: #181

### DIFF
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -311,13 +311,6 @@ class Consumer(object):
             logger.error('No commit id in message!')
             return
 
-        # Check the login. If it's bad then a koji.AuthError exception will get thrown.
-        # and we'll exit because catch_exceptions_and_continue() will raise it.
-        # This is the only way we've come up with to deal with the problem in
-        # https://github.com/coreos/fedora-coreos-releng-automation/issues/70
-        if self.keytab_file:
-            self.koji_client.getLoggedInUser()
-
         self.process_git_lockfiles(commit)
 
     @catch_exceptions_and_continue
@@ -353,6 +346,13 @@ class Consumer(object):
         self.tag_rpms(desiredrpms)
 
     def tag_rpms(self, desiredrpms):
+
+        # Check the login. If it's bad then a koji.AuthError exception will get thrown.
+        # and we'll exit because catch_exceptions_and_continue() will raise it.
+        # This is the only way we've come up with to deal with the problem in
+        # https://github.com/coreos/fedora-coreos-releng-automation/issues/70
+        if self.keytab_file:
+            self.koji_client.getLoggedInUser()
 
         # NOMENCLATURE:
         # 
--- a/coreos-koji-tagger/coreos_koji_tagger.py
+++ b/coreos-koji-tagger/coreos_koji_tagger.py
@@ -186,6 +186,16 @@ EXAMPLE_GITHUB_PUSH_MESSAGE_BODY = json.loads("""
 """
 )
 
+# We also process the org.fedoraproject.prod.coreos.stream.release topic
+# https://apps.fedoraproject.org/datagrepper/raw?topic=org.fedoraproject.prod.coreos.stream.release&delta=100000
+EXAMPLE_COREOS_STREAM_RELEASE_MESSAGE_BODY = json.loads("""
+{
+    "basearches": ["x86_64", "s390x", "ppc64le"],
+    "build_id": "40.20240113.91.1",
+    "stream": "rawhide"
+}
+""")
+
 def catch_exceptions_and_continue(func):
     # This is a decorator function that will re-call the decorated
     # function and will catch any exceptions and not raise them further.
@@ -282,6 +292,8 @@ class Consumer(object):
 
         if message.topic == 'org.fedoraproject.prod.github.push':
             self.process_github_push_message(message)
+        elif message.topic == 'org.fedoraproject.prod.coreos.stream.release':
+            self.process_coreos_stream_release_message(message)
         else:
             # that's weird, we shouldn't have been called for any other topic
             # just log it and ignore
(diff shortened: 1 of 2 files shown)


coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -341,6 +351,11 @@ def archify(f):
             logger.warning('Continuing...')
             return
 
+        self.tag_rpms(desiredrpms)
+
+    @catch_exceptions_and_continue
```
dustymabe, line 378: I don't think we need the `@catch_exceptions_and_continue` at the top of all of these functions do we? In the original file there were only two.. one for the "initial run" path and one for the "processing messages" path. If a function is called from another function that has `@catch_exceptions_and_continue` then it will be caught at the top level.

coreos-koji-tagger/coreos_koji_tagger.py
```diff
@@ -314,6 +326,23 @@ def process_github_push_message(self, message: fedora_messaging.api.Message):
...
         self.process_git_lockfiles(commit)
 
+    @catch_exceptions_and_continue
+    def process_coreos_stream_release_message(self, message: fedora_messaging.api.Message):
+        msg = message.body
+        stream = msg['stream']
+
+        # we only handle rawhide and branched builds; see
+        # https://github.com/coreos/fedora-coreos-releng-automation/issues/181
+        if stream not in ['rawhide', 'branched']:
```
dustymabe, line 336: Operating on `rawhide` now might mean we need to change our SOP for updating the signatures for our pool. See https://github.com/coreos/fedora-coreos-tracker/blob/fe29ffc877c504482dc9bd6e17a1ed9a79eeb628/.github/ISSUE_TEMPLATE/rebase.md?plain=1#L30-L32

dustymabe: We also may want to be more aggressive about pruning rawhide rpms since this will cause the size of our pool to rise dramatically (IMO), which will have a side effect of causing the distrepo tasks to take longer.

jlebon: > Operating on `rawhide` now might mean we need to change our SOP for updating the signatures for our pool. See [coreos/fedora-coreos-tracker@`fe29ffc`/.github/ISSUE_TEMPLATE/rebase.md?plain=1#L30-L32](https://github.com/coreos/fedora-coreos-tracker/blob/fe29ffc877c504482dc9bd6e17a1ed9a79eeb628/.github/ISSUE_TEMPLATE/rebase.md?plain=1#L30-L32)

Isn't that command already doing what we need? (It's adding the rawhide key after rawhide moves to N+1). E.g. right now, the coreos-pool tag has the rawhide key in its `tag2distrepo.keys`.

> We also may want to be more aggressive about pruning rawhide rpms since this will cause the size of our pool to rise dramatically (IMO), which will have a side effect of causing the distrepo tasks to take longer.

Yes, definitely. I think one thing we can do here is at branching time, we reset everything: for Bodhi testing of rawhide, we don't want to use a e.g. f40 lockfile to test an f41 package since it might not even depsolve, and for Bodhi testing of branched, I think it'll be too complex to have the code know to take the last successful rawhide build and test from that, so it'd make more sense to just not use the "auto-lockfile" behaviour there until we get a clean branched build to base of off going forward.

coreos-koji-tagger/coreos_koji_tagger.py
jlebon: Yes, absolutely. I was incorrectly cargo culting there.

jlebon:
Updated!

coreos-koji-tagger/coreos_koji_tagger.py
dustymabe: > > Operating on `rawhide` now might mean we need to change our SOP for updating the signatures for our pool. See [coreos/fedora-coreos-tracker@`fe29ffc`/.github/ISSUE_TEMPLATE/rebase.md?plain=1#L30-L32](https://github.com/coreos/fedora-coreos-tracker/blob/fe29ffc877c504482dc9bd6e17a1ed9a79eeb628/.github/ISSUE_TEMPLATE/rebase.md?plain=1#L30-L32)
> 
> Isn't that command already doing what we need? (It's adding the rawhide key after rawhide moves to N+1). E.g. right now, the coreos-pool tag has the rawhide key in its `tag2distrepo.keys`.

Ahh yes.. I forgot we moved it earlier and it covers rawhide now..  Ignore me.

> 
> > We also may want to be more aggressive about pruning rawhide rpms since this will cause the size of our pool to rise dramatically (IMO), which will have a side effect of causing the distrepo tasks to take longer.
> 
> Yes, definitely. I think one thing we can do here is at branching time, we reset everything: for Bodhi testing of rawhide, we don't want to use a e.g. f40 lockfile to test an f41 package since it might not even depsolve, and for Bodhi testing of branched, I think it'll be too complex to have the code know to take the last successful rawhide build and test from that, so it'd make more sense to just not use the "auto-lockfile" behaviour there until we get a clean branched build to base of off going forward.

I think ther are some details here I'm not quite fully groking but we can dig into that later.

jlebon: Sorry, "we reset everything" is quite vague. To be more specific, I mean "we untag all rawhide packages".

dustymabe (approved):
LGTM - merge when ready

jlebon:
This looks like it's working well in production!

One thing this is missing is syncing on start like the git lockfiles logic. We can add that in a follow-up eventually but for now it also works to just re-trigger a release job to get the message sent and trigger the logic here.

PR #184, opened 2024-02-12
Dockerfile: Move to Fedora 39

c4rt0:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1490

### DIFF
--- a/config-bot/Dockerfile
+++ b/config-bot/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:38
+FROM registry.fedoraproject.org/fedora:39
 
 RUN dnf -y install git python3-aiohttp && dnf clean all
 


jlebon (approved):
Thanks! Ideally we'd actually test this before merging, but in the past it's been quite safe. When merging this (and other PRs like this one that update a service), watch the rebuild and deployment in OpenShift to sanity-check it comes up fine.

c4rt0:
> When merging this (and other PRs like this one that update a service), watch the rebuild and deployment in OpenShift to sanity-check it comes up fine.

Will keep that in mind, thanks for the  review!

dustymabe (approved):

c4rt0:
I found this on Openshift (no errors were recorded):

```
STEP 7/7: LABEL
...
 "io.openshift.build.commit.message"="Merge pull request #186 from c4rt0/moveToFedora39"
  "io.openshift.build.commit.ref"="main" "io.openshift.build.name"="config-bot-39"  
```
I guess that's safe.

jlebon:
Yeah, that's the image build. You'll want to also check that the new pod comes up and is happy (`oc get pods`).

Also, this would happen after that PR is merged. Those logs are from when https://github.com/coreos/fedora-coreos-releng-automation/pull/186 was merged it looks like. (But a good time to also check that the importer is also up and running fine.)

PR #185, opened 2024-02-12
coreos-koji-tagger/Dockerfile: move to Fedora 39

c4rt0:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1490


dustymabe (approved):
LGTM

PR #190, opened 2024-11-25
scripts/promote-config: handle git submodules

jlebon:
This script isn't quite doing the right thing now that we've added the base-images repo as a submodule. We want to treat the git submodule just like any other file; updates to it should also be promoted.

Tweak the `git reset` to make this happen. Squash a warning that `git add -A` emits about this since it's expected.


dustymabe (approved):
LGTM

PR #193, opened 2024-12-10
fedora-ostree-pruner: do repo checks on initial startup too

dustymabe:
This will make sure that messages are in the logs and also I think now that we have proper alerting we'll get an alert when pods exit because of an Exception like this.


marmijo (approved):
LGTM

PR #194, opened 2024-12-18
scripts/promote-config: fix git submodule handling

jlebon:
The `--recurse-submodules` option to `git reset` only works if there are already submodules to recurse into. In the case where we're promoting from one branch which has a submodule to one which doesn't, this will not cause a git submodule directory to be created. Instead, we need to explicitly initialize potential new submodules.

We also need to handle the case where the submodule did exist but it was bumped by the promotion and so we need to update it.

Calling `git submodule update --init` handles both cases.

### DIFF
--- a/scripts/promote-config.sh
+++ b/scripts/promote-config.sh
@@ -30,7 +30,8 @@ main() {
     head=$(git rev-parse HEAD)
 
     # take all the changes from the src branch, including any submodules
-    git reset --hard "${fetch_head}" --recurse-submodules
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

dustymabe:
after this merges we can add the `ok-to-promote` label on https://github.com/coreos/fedora-coreos-streams/issues/1002 to test it - we can then close the PR since we don't want to promote yet.

PR #199, opened 2025-03-05
coreos-koji-tagger: log when no ref in msg

dustymabe:
Been seeing a lot of `KeyError: 'ref'` lately so let's add some debugging to try to figure it out.


marmijo (approved):
LGTM

PR #203, opened 2025-04-14
coreos-koji-tagger: bump signing timeout to 90m

dustymabe:
Sometimes builds get stuck in the signer and the signer itself has a timeout at 60 minutes. So if we get in this situation we could have to wait 60 minutes and then also for the queue to get processed down to our position in the queue. Let's try 90 minutes as a timeout to see if we can get less runs that give up before getting to the distrepo step.


marmijo (approved):
LGTM

aaradhak (approved):
LGTM

PR #210, opened 2025-06-26
coreos-koji-tagger/Dockerfile: create the my_config.toml file

dustymabe:
The line that was deleted in the last commit actually did two things. It rand a find/replace and also created the file. Let's do the creation of the file in the new first sed command.

Fixes 16e368f

### DIFF
--- a/coreos-koji-tagger/Dockerfile
+++ b/coreos-koji-tagger/Dockerfile
@@ -23,7 +23,7 @@ RUN mkdir /work
 WORKDIR /work
 
 # Set the Application Name
-RUN sed -i 's|Example Application|CoreOS Koji Tagger: https://github.com/coreos/fedora-coreos-releng-automation/tree/main/coreos-koji-tagger |' /work/my_config.toml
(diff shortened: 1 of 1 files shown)


c4rt0:
Sorry for my curiosity here @dustymabe, I was reviewing this and I see that `sed -i` is used to edit files in-place instead of printing to standard output... and I also see the [redirection](https://github.com/coreos/fedora-coreos-releng-automation/pull/210/files#diff-6e571b565a1b6804a1db474d6287fd318f99482a0493d13ae9f02b78845f1847R26)

https://www.gnu.org/software/sed/manual/sed.html#The-_002di-option 

I know this is merged already, just wondering if removing the `-i` wouldn't be a good idea?

dustymabe:
> Sorry for my curiosity here @dustymabe, I was reviewing this and I see that `sed -i` is used to edit files in-place instead of printing to standard output... and I also see the [redirection](https://github.com/coreos/fedora-coreos-releng-automation/pull/210/files#diff-6e571b565a1b6804a1db474d6287fd318f99482a0493d13ae9f02b78845f1847R26)
> 
> https://www.gnu.org/software/sed/manual/sed.html#The-_002di-option
> 
> I know this is merged already, just wondering if removing the `-i` wouldn't be a good idea?

Precisely! Good Catch! https://github.com/coreos/fedora-coreos-releng-automation/pull/211

PR #212, opened 2025-07-09
config-bot: update apiVersion as required by OpenShift

dustymabe:
Newer OpenShift will no longer process templates without being more verbose with the specification.


yasminvalim (approved):
/lgtm

PR #229, opened 2026-04-01
Move 41 to EOL and add 44 to stable

Zlopez:
The missing policies for F44 is causing OpenShift pod to Crashloop.

This change should fix it.


Zlopez:
I wonder if it wouldn't be better for future to just move the global variables to external config, which could be easily changed in ansible.

dustymabe:
> I wonder if it wouldn't be better for future to just move the global variables to external config, which could be easily changed in ansible.

ehh. If there were more than one instance of this thing running somewhere then yeah. but at this point it's still just a git commit and having the config live with the code decreases complexity IMO.

dustymabe (approved):
LGTM

Zlopez:
Thanks, let me redeploy it in OpenShift.

dustymabe:
> Thanks, let me redeploy it in OpenShift.

builds get autotriggered

PR #232, opened 2026-08-18
fedora-ostree-pruner: add 45 to the stable list, drop 42

samyak-jn:
F45 Atomic Desktop refs are already in /mnt/koji/ostree/repo after mass branching, but the pruner still only knows about 42–44, so it fails with "Found refs with no policy defined."


Zlopez (approved):

dustymabe (approved):
LGTM

